PR #57, opened 2016-05-23
Add base for custom web app

msmolens:
This replaces the current Girder web client extensions with a minimal custom web app. The app has a header bar, a front page, an example view, and now a mockup for an upload dataset view.

### DIFF
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -4,16 +4,49 @@
 import os
 
 from girder import events
-from girder.constants import SettingKey
+from girder.constants import SettingKey, PACKAGE_DIR, STATIC_ROOT_DIR
 from girder.models.model_base import ValidationException
 from girder.utility.model_importer import ModelImporter
 from girder.utility.server import staticFile
+from girder.utility.webroot import WebrootBase
 
 from . import constants
 from . import api
 from .provision_utility import initialSetup, onUserCreated
 from .task_utility import UDAResource, TaskHandler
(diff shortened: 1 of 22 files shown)


brianhelba:
Let's change the namespace to `isic`.

brianhelba:
Drop the upload dataset view functionality from this PR please. We'll do that separately.

server/__init__.py
```diff
@@ -98,3 +131,8 @@ class Root(object):
     info['apiRoot'].image = api.ImageResource()
     info['apiRoot'].segmentation = api.SegmentationResource()
     info['apiRoot'].study = api.StudyResource()
+
+    # Move girder app to /girder, serve isic_archive app from /
+    info['serverRoot'], info['serverRoot'].girder = (Webroot(),
```
brianhelba: Let's keep the Girder app mounted at `/`. Mount this new app at `/isic` for now.

server/__init__.py
```diff
@@ -98,3 +131,8 @@ class Root(object):
...
+                                                     info['serverRoot'])
+    info['serverRoot'].api = info['serverRoot'].girder.api
```
brianhelba: We're not ready to remap the API yet, since we still have Angular code where it's hardcoded. Comment out this line.

msmolens:
@brianhelba PTAL

brianhelba:
LGTM.

PR #59, opened 2016-06-03
Update client-side check whether user can contribute datasets

msmolens:
The client now checks whether the current user is a member of the dataset
contributors group. This determines whether the 'Upload dataset' link is shown
on the front page and whether the upload dataset view route can be accessed.


brianhelba:
LGTM.

PR #60, opened 2016-06-03
Remove hardcoded relative path from ESLint config

msmolens:
This commit allows running the ESLint JavaScript static analysis tests no matter
where the repository is located relative to the Girder repository. For example,
the tests now run when the plugin is deployed using Vagrant.


brianhelba:
LGTM.

PR #61, opened 2016-06-07
WIP: Add study page

msmolens:
The study page lists annotation studies. Initially, each study's name is shown
in the header of a collapsed panel. When the user clicks a study's name, the
panel expands to show the study details. Currently, only the information
directly available from the /study/id/* endpoints is fetched/shown.

### DIFF
--- a/web_external/js/views/body/StudyView.js
+++ b/web_external/js/views/body/StudyView.js
@@ -1,0 +1,62 @@
+isic.views.StudyView = isic.View.extend({
+    initialize: function (settings) {
+        girder.cancelRestRequests('fetch');
+
+        this.study = settings.study;
+
+        var promises = [];
+
+        // TODO: Add way to detect collection fetch error. On error, set
+        // collection to null and reject promise.
+
+        var usersDeferred = $.Deferred();
+        promises.push(usersDeferred.promise());
+        this.users = new girder.collections.UserCollection();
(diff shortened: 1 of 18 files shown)


brianhelba:
@msmolens I think this is good enough to merge. PTAL.

Let's add additional features as additional PRs, to keep things from piling up.

web_external/js/views/body/StudyView.js
```diff
@@ -0,0 +1,26 @@
+isic.views.StudyView = isic.View.extend({
+    initialize: function (settings) {
+        girder.cancelRestRequests('fetch');
+
+        if (settings.study) {
+            this.study = settings.study;
+            this.render();
+        } else if (settings.id) {
```
brianhelba: Generally, I think a View should be responsible for fetching its own Model with details. However, I've left it an option to pass in a pre-fetched Model.

@msmolens What do you think of this policy in general?

brianhelba:
Merging this now, but please let me know your feedback, @msmolens .

web_external/js/views/body/StudyView.js
```diff
@@ -0,0 +1,26 @@
+isic.views.StudyView = isic.View.extend({
+    initialize: function (settings) {
+        girder.cancelRestRequests('fetch');
```
brianhelba: @msmolens Does this cancel all requests globally? If so, what if we're loading 2 StudyViews concurrently?

PR #63, opened 2016-06-14
Add header buttons to standalone web app

brianhelba:


brianhelba:
The upload dataset button still needs to be selectively displayed, given a user's permissions for it.

brianhelba:
@msmolens PTAL

One bug is that the "Upload Dataset" button on the header is not being removed after the user logs out (though it is added when an authorized user logs in). Feel free to add a commit to fix this, and merge if you think the code is good otherwise.

msmolens:
@brianhelba Your commits LGTM. PTAL at what I added.

brianhelba:
LGTM.

PR #64, opened 2016-06-14
Ansible tweaks

msmolens:
This addresses a few problems that I hit while attempting to provision on Windows with Vagrant 1.8.4.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -31,7 +31,7 @@ Vagrant.configure("2") do |config|
     # ansible.raw_arguments = ["--module-path=" + File.expand_path("ansible/library")]
     ENV["ANSIBLE_LIBRARY"] = File.expand_path("ansible/library")
     if provisioner_type == "ansible_local"
-      ansible.provisioning_path = "/home/vagrant/digital_slide_archive"
+      ansible.provisioning_path = "/home/vagrant/isic_archive"
     end
   end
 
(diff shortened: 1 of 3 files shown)


brianhelba:
@msmolens Thanks. This fixes a few obvious bugs due to copy-pasted code and a lack of testing on my part.

See my inline comment for 1 issue.

Vagrantfile
```diff
@@ -27,11 +27,8 @@ Vagrant.configure("2") do |config|
   config.vm.provision provisioner_type do |ansible|
     ansible.playbook = "ansible/vagrant-playbook.yml"
     ansible.galaxy_role_file = "ansible/requirements.yml"
-    # Ansible has a bug where the "--module-path" option is not respected
-    # ansible.raw_arguments = ["--module-path=" + File.expand_path("ansible/library")]
-    ENV["ANSIBLE_LIBRARY"] = File.expand_path("ansible/library")
```
brianhelba: Let's leave these lines about setting the `ENV`, as it's still necessary for ansible running on the host. The config file points to a path that only exists within the guest.

ansible.cfg
```diff
@@ -0,0 +1,3 @@
+[defaults]
+
+library = /home/vagrant/isic_archive/ansible/library
```
brianhelba: Eventually, I'll get Conda support pushed upstream to Ansible, and this won't be necessary anymore. Good workaround for now.

Vagrantfile
msmolens: Makes sense. I updated the commit.

PR #65, opened 2016-06-15
Expand study API

msmolens:


server/api/study.py
```diff
@@ -243,30 +245,87 @@ def redirectTask(self, study, params):
...
+        .param('id', 'The ID of the study.', paramType='path')
+        .param('userId', 'The ID of the user.')
+        .errorResponse('ID was invalid.')
+        .errorResponse('You don\'t have permission to add a study annotator.', 403)
+    )
+    @access.user
+    @loadmodel(model='study', plugin='isic_archive', level=AccessType.WRITE)
+    def addAnnotator(self, study, params):
+        self.requireParams('userId', params)
+        user = self.getCurrentUser()
```
brianhelba: Let's call this `creator_user` for clarity.

brianhelba:
I added some new commits to clean up the code. I didn't want to trample your development, but it was easier to just make the changes myself than describe them.

PTAL.

server/api/study.py
```diff
@@ -255,58 +255,45 @@ def _requireStudyCreator(self, user):
...
-            study = self.model('study', 'isic_archive').createStudy(
-                name, user, featureset, annotator_users, segmentations)
-        else:
-            # Use query parameters
-            self.requireParams(('name', 'featuresetId'), params)
-            user = self.getCurrentUser()
-            self._requireStudyCreator(user)
-
-            name = params['name'].strip()
+        if cherrypy.request.headers['Content-Type'] == 'application/json':
```
msmolens: The reason I didn't check the header like this is that calling this endpoint through Swagger sends `Content-Type:application/json`, and the response is then:

    {
      "message": "Invalid JSON passed in request body.",
      "type": "rest"
    }

msmolens:
@brianhelba PTAL

brianhelba:
This looks good to go. Merging.

PR #67, opened 2016-06-20
Add a featureset listing page

brianhelba:


brianhelba:
I've added the API changes that should be necessary for this. Now we just need the client-side work.

I'll add the support for searching featuresets by name in a separate PR, as that's not needed for the listing.

msmolens:
@brianhelba PTAL. The views could be refactored more when we have time.

brianhelba:
@msmolens Nice job with the listingPage. This should make future development much easier.

At some point we'll want to have the featureset name in the study detail hyperlink to the corresponding featureset detail. However, that will require us to add routing to each expanded featureset detail (or more generally, each expanded listing detail). I don't think this is helpful enough to bother with quite yet.

LGTM. Merging this.

PR #68, opened 2016-06-21
Add a create study UI

brianhelba:


brianhelba:
I'll add server-side support for full-text searching in a bit, but this requires a bit more messing with DB indexes.

msmolens:
Investigating best solution for a bug when searching on Models that aren't AccessControlledModels:

    TypeError: TypeError("prefixSearch() got an unexpected keyword argument 'user'")

msmolens:
@brianhelba Right now this includes user selection and featureset selection, but not image selection. Depends on https://github.com/girder/girder/pull/1405.

brianhelba:
Merging now, to prevent this from being an outstanding branch for too long. We can add image selection later.

PR #70, opened 2016-06-22
Allow the magic "filename" field in CSV files to be case-insensitive

brianhelba:


brianhelba:
@msmolens I'm merging now, but PTAL, since you've touched this code before.

msmolens:
LGTM

PR #75, opened 2016-08-10
Update the front page of standalone web app

aadikalloo:


web_external/templates/body/frontPage.jade
```diff
@@ -1,56 +1,102 @@
...
+.isic-body-nopad.isic-frontpage
+  .isic-frontpage-header
+    .container
+      h1 ISIC Archive
+      h2 International Skin Imaging Collaboration: Melanoma Project
+  .isic-frontpage-body.container
+    .row
+      .col-lg-6
+        h3 Introduction
+        p
```
msmolens: It could be cleaner to write these paragraphs using `p.` instead of prefixing each line with `|`.
See http://jade-lang.com/reference/plain-text/, ("Block in a Tag")

brianhelba:
LGTM. Merging.

PR #93, opened 2016-08-16
Add annotation study results page

msmolens:

PR #94, opened 2016-08-16
Misc. API improvements

brianhelba:


brianhelba:
@msmolens FYI, `GET /study/{id}` now returns an `images` field, which should be used for #93 . `segmentations` will be removed from that endpoint soon.

PR #96, opened 2016-08-17
Remove Phase 1

brianhelba:


brianhelba:
@msmolens PTAL (no need to review every line though). I've got a little more testing and polishing to do, but then this should be enough to support moving all images in Phase 1a and 1b into Lesion Images without any loss of functionality for segmentation.

msmolens:
@brianhelba LGTM so far

PR #100, opened 2016-08-18
WIP: Add WorkingGroupMembers Page

aadikalloo:
A page that lists the different working groups of ISIC and the members
who belong to each group. Would benefit from being a dropdown option from About.
@brianhelba I quickly copied the HTML. The original page is here: [http://isdis.net/isic-project/working-group-members/](http://isdis.net/isic-project/working-group-members/). The existing isic stylesheets are fine for this


brianhelba:
@aadikalloo I'm going to close this, since we now have the external link. Ok?

aadikalloo:
yes

PR #104, opened 2016-08-19
Images gallery serverside histograms

alex-r-bigelow:
To test, hit:
[isic-archive.dev/api/v1/image/histogram](http://isic-archive.dev/api/v1/image/histogram)


alex-r-bigelow:
Sorry, I just force-pushed again—now this branch should be clean and ready

web_external/js/.eslintrc.js
```diff
@@ -1,7 +1,7 @@
 var path = require('path');
 
 module.exports = {
-    "extends": path.join(process.cwd(), ".eslintrc"),
+    //"extends": path.join(process.cwd(), ".eslintrc"),
```
brianhelba: Revert this.

server/histogram_utility/__init__.py
```diff
@@ -0,0 +1,204 @@
...
+TRUE_VALUES = set([True, 'true', 1, 'True'])
+
+
+class HistogramUtility():
+    def __init__(self):
+        # Load up the external foreign code snippets
+        self.foreignCode = {}
+
+        codePath = 'plugins/isic_archive/server/histogram_utility'
+        codePath = os.path.join(os.getcwd(), codePath)
```
brianhelba: Can we use `__file__`?

server/histogram_utility/__init__.py
```diff
@@ -0,0 +1,204 @@
...
+
+        for filename in os.listdir(codePath):
+            extension = os.path.splitext(filename)[1]
+            if extension != '.js' and extension != '.json':
+                continue
+            infile = open(os.path.join(codePath, filename), 'rb')
```
brianhelba: Should use a `with` statement to open.

server/api/image.py
```diff
@@ -147,6 +151,28 @@ def download(self, image, params):
...
+        .param('limit', 'Result set size limit. Setting to 0 will create ' +
+               'a histogram using all the matching items (default=0).',
+               required=False, dataType='int')
+        .param('offset', 'Offset into result set (default=0).',
+               required=False, dataType='int')
+        .errorResponse()
+    )
+    @access.public
+    def histogram(self, params):
+        # TODO: ideally we should access images via the image route, but it
```
brianhelba: Because `Image` is a subtype of `Item`, you can get all Images with:
```
imagesCollection = self.model('collection').collection.find_one({'name': 'Lesion Images'})
allImagesCursor = self.model('item').collection.find({'baseParentId': imagesCollection['_id']})
```
Of course, just calling `self.model('image', 'isic_archive').find()` does all this for you.

Is there any way to pre-filter things here, and pass the Images cursor into `histogramUtility.getHistograms`, instead of passing an unfiltered collection?

alex-r-bigelow: AFAIK, we need the collection to run mapreduce code.

We *do* [prefilter](https://github.com/ImageMarkup/isic-archive/blob/images-gallery-serverside-histograms/server/histogram_utility/__init__.py#L33) the collection before running mapreduce—it's kind of an optimized, all-in-one call.

alex-r-bigelow:
@brianhelba It should be ready now—though just to make sure:
```python
Dataset.list(user=user)
```
returns folders that only contain Image items, right? Could there be any non-image items in those folders? If so, I still need to add another filter.

PR #105, opened 2016-08-19
Images gallery move histogram

alex-r-bigelow:
- [x] Move the histogram to the bottom-left
- [x] Remove the toggle button
- [x] Add a close box to the image details pane


brianhelba:
@alex-r-bigelow From your commit, I see that you're using `var self = this;` in some places. Can we phase this out, in favor of using [`_.bind`](http://underscorejs.org/#bind), as we do in the rest of Girder?

brianhelba:
It looks like all of these commits are included in https://github.com/ImageMarkup/isic-archive/pull/109 , so I'm closing this PR.

PR #109, opened 2016-08-22
Images gallery pagination

alex-r-bigelow:
Still need to:
- [x] Display text along the lines of "Displaying images 1 - 50 of 500 filtered images"
- [x] Show a bar indicating the same information

### DIFF
--- a/web_external/js/views/body/ImagesView/HistogramPane.js
+++ b/web_external/js/views/body/ImagesView/HistogramPane.js
@@ -3,8 +3,9 @@
 isic.views.ImagesSubViews = isic.views.ImagesSubViews || {};
 
 isic.views.ImagesSubViews.HistogramPane = Backbone.View.extend({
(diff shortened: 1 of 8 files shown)


brianhelba:
@alex-r-bigelow This is now pulling in all of the server-side changes in `images-gallery-serverside-histograms`. I'd like to review those as a separate change, but I remember we discussed that the pagination might need it. Should we wait on merging this, and review / merge #104 next now?

alex-r-bigelow:
The server-side changes are necessary for the "pagination" feature to be complete, as we discussed (e.g. the seek to end button, among other things, will not work without additional information from the server).

I think the correct order would be to review / merge #104 first.

alex-r-bigelow:
@brianhelba I'm not going to push to this branch anymore, so feel free to commit if you see things that you want to change.

PR #112, opened 2016-08-23
Images gallery histograms

brianhelba:


brianhelba:
@alex-r-bigelow I've rebased your code in the branch `images-gallery-histograms` on top of the latest `master`. Please let me me know if you've got anything locally that wasn't pushed and you need to rebase on top of this further, as I can to help you get that done quickly.

As far as I'm aware, this branch / PR is the only thing still outstanding from you. Everything else should now be merged.

alex-r-bigelow:
Everything looks like it's in order, thanks! I had some stuff that I hadn't committed yet, so we shouldn't need to rebase.

alex-r-bigelow:
@brianhelba To get filtering to work, I'll need to install peg.js; I assume the best way to do this would be to add a line to bower.json?

brianhelba:
Add it to`package.json`.

brianhelba:
@msmolens Can you get `pegjs` to fetch from NPM and build into the plugin JS?

msmolens:
@alex-r-bigelow What error did you get when adding `pegjs` to `package.json`? I get `EPROTO: protocol error, symlink...` during `npm install`, but that's expected on Windows with the current configuration.

alex-r-bigelow:
I don't get any errors; I'm expecting the `peg` variable (I guess, more precisely, `window.peg`) to exist, but every configuration that I tried didn't seem to quite get there... I'm sure there's something simple I was missing.

alex-r-bigelow:
I can't seem to crack the filtering problems—the backbone events seem fixed, but the wrong bars render (only the page histogram seems to update—the filter bars don't), and the image wall doesn't seem to be getting the event indicating that the filtered set has changed.

There *should* be enough of a response now to chase down the bug... but for clean merging / demonstration purposes, I've hidden the filtering checkmarks altogether (see the last commit).

I'll go ahead and remove the WIP, and the filter issues can be a separate branch/PR.

brianhelba:
@alex-r-bigelow Ok, got it.

I'm adding the `pegjs` package to the build system now. What version do you want? Your commit to `package.json` uses `0.9.0`, but your inlined file is `0.10.0`.

alex-r-bigelow:
We used 0.9.0 in Resonant Lab, so I'd default to that (@ronichoudhury knows more about the pegjs stuff than I do)

brianhelba:
Ok @ronichoudhury , I'm going to specify:
```
"pegjs": "~0.9.0"
```

PR #117, opened 2016-08-24
Implement tasks page using Backbone

### LINKED ISSUE #110: Refactor tasks page to use Backbone
The page should display tasks for 
- `Phase 0`
  - Subtitle: `Images to QC`
  - Call `/api/v1/uda/task/qc` for tasks
  - Redirect each task to `/uda/task/p0/#{task.dataset._id}`
- `Lesion Segmentation`
  - Subtitle: `Segment boundaries between lesion and normal skin`
  - Call `/api/v1/task/me/segmentation` for tasks
  - Redirect each task to `/api/v1/task/me/segmentation/redirect?datasetId=#{task.dataset._id }`
- `Annotation Studies`
  - Subtitle: `Clinical feature annotation studies`
  - `Call `/api/v1/task/me/annotation` for tasks`
  - Redirect each task to `/api/v1/task/me/annotation/redirect?studyId=#{task.study._id }`

Each task, category, and the total should include badges displaying the task counts.

Also, [these YouTube videos should be embedded](https://github.com/ImageMarkup/isic-archive/blob/8b0168b6d98df267e6cf2538e1274ba461931ee1/custom/task.html#L126), though they may be removed eventually.

We will eventually add some logic to only fetch some of these task categories / API endpoints when the current user has the appropriate permissions / group memberships, but this isn't necessary immediately.

msmolens:
Fixes #110


brianhelba:
@msmolens The old page did an API call periodically to check for new data. I don't think this is that necessary, and tends to load our server when people leave the page idle.

However, to make up for this, can you add a button with a refresh icon at the top of the page (to the right of the "Task Dashboard" title)? When clicked, it should trigger a re-fetch (and event-driven re-render when the fetch completes) of all 3 `TaskCollection`s.

web_external/templates/body/tasksPage.jade
```diff
@@ -0,0 +1,15 @@
+.container.isic-tasks-container
```
brianhelba: @msmolens Following with what our other views do, and to keep the proliferation of stylesheets under control, I think we can put all this in `taskPage.styl`, right?

brianhelba: Oops, I meant the other way around: put the content of `taskPage.styl` under `taskPage.styl`.

msmolens:
@brianhelba Yes, that's cleaner.

PR #120, opened 2016-08-25
Lesion images: Add image details

msmolens:


web_external/templates/widgets/imageDetailsPage.jade
```diff
@@ -0,0 +1,10 @@
+.isic-image-details-container
+  .button
+    img(src='#{imgRoot}/ex.svg')
+  if image.id
+    .isic-image-details-name= image.name()
+    .isic-image-details-clinical-metadata
```
brianhelba: Shouldn't there be a `dl` here?

PR #127, opened 2016-08-29
Study Results tweaks

msmolens:
- [x] fix error when viewing certain images
- [x] grey out global feature id when there is no value
- [x] disable local feature items when selected annotation has no data
- [ ] <strike>[check whether annotation is complete?</strike> -- not now


msmolens:
@brianhelba PTAL

brianhelba:
LGTM.

PR #132, opened 2016-08-30
Display Featureset in a nicer table

brianhelba:


brianhelba:
@msmolens PTAL.

If you have any ideas to improve the styling of these tables, please feel free to implement them.

PR #140, opened 2016-08-31
Update to newer version of Girder

ronichoudhury:
attn: @brianhelba

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ addons:
             - p7zip-full
 
 before_install:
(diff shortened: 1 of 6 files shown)


msmolens:
Should Travis CI use the same version? See https://github.com/ImageMarkup/isic-archive/blob/ca3ad6f8f1fefe4ba237016f48e54fd53b66c2aa/.travis.yml#L35

brianhelba:
Let me take some time to review some of the upstream diffs between this upgrade and the previously-used version. Then I'll switch Travis and the live deployment to all use the same version.

PR #141, opened 2016-08-31
Reformat featureset table

msmolens:
Perhaps it's an improvement?

### DIFF
--- a/web_external/stylesheets/body/featuresetPage.styl
+++ b/web_external/stylesheets/body/featuresetPage.styl
@@ -15,7 +15,7 @@
     .isic-featureset-table-feature-header
       border 0px
       padding-top 10px
-      font-size 20px
+      font-size 16px
 
     .isic-featureset-table-feature-id-key
     .isic-featureset-table-feature-id-value
--- a/web_external/stylesheets/body/featuresetPage.styl
+++ b/web_external/stylesheets/body/featuresetPage.styl
@@ -1,4 +1,40 @@
-.isic-featureset-global-features-content
-.isic-featureset-local-features-content
-  display inline-block
-  width 800px
+.isic-featureset-container
+
+  .isic-featureset-table
+    width 100%
+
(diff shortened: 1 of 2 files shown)


web_external/templates/body/featuresetPage.jade
```diff
@@ -1,54 +1,47 @@
...
+        td(colspan=2)
+          if featureset.get(featureType.id).length
+            table.table.table-condensed.isic-featureset-table-features-table
+              each feature in featureset.get(featureType.id)
                 tr
-                  th ID
-                  th Name
-                each valueOption in feature.options
+                  td.isic-featureset-table-feature-header(colspan=3)= feature.name.join(', ')
+                    | :
```
brianhelba: I don't think we want this `:`, since the change in font already differentiates the name from the rest.

web_external/stylesheets/body/featuresetPage.styl
```diff
@@ -1,4 +1,40 @@
...
+
+    .isic-featureset-table-section-header
+      background-color #eee
+      padding 0px 8px
+      font-size 20px
+
+    .isic-featureset-table-feature-header
+      border 0px
+      padding-top 10px
+      font-size 20px
```
brianhelba: This still appears slightly emphasized at `16px`, and takes up a lot less space.

web_external/stylesheets/body/featuresetPage.styl
```diff
@@ -1,4 +1,40 @@
-.isic-featureset-global-features-content
-.isic-featureset-local-features-content
-  display inline-block
-  width 800px
+.isic-featureset-container
+
+  .isic-featureset-table
+    width 100%
```
brianhelba: I don't think it looks good to have empty table stretching so far to the right, but maybe the solution is to make all of the `listingPage` occupy less than the full width of the page (which would better match the front page).

PR #144, opened 2016-08-31
Use client-side utility function to format users as strings

msmolens:
This handles formatting users on the client once the code in https://github.com/ImageMarkup/isic-archive/blob/e892e7e248773c8ef346cc5b82bc794833eae9bb/server/models/user.py#L36 is activated,


brianhelba:
The server-side code is now active. Instead of just determining whether the current user is a site admin, the permissions for the current user viewing the particular user are checked.

I'll add the functionality for anonymous studies later.

PR #146, opened 2016-08-31
Fix disappearing paging buttons

### LINKED ISSUE #134: Image gallery: SVG forward/backward page buttons disappear on mouseover

ronichoudhury:
Firefox seems to have trouble with the advanced SVG magic being used to recolor the images.

Instead of all that, this PR just uses opacity as a way to signal the (1) disabled state of a paging button and (2) rollover state of the same.

Fixes #134

### DIFF
--- a/web_external/stylesheets/body/imagesPage.styl
+++ b/web_external/stylesheets/body/imagesPage.styl
@@ -56,6 +56,7 @@ $dataAxisColor = #333333
         img
             width 1em
             height 1em
+            opacity 0.7
 
         &.selected
             img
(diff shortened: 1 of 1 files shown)


brianhelba:
Are the `rolloverImageFilter` and `disabledImageFilter` now unused? If so, can you delete the dead code?

brianhelba:
@ronichoudhury If not, should we replace those filters with CSS opacity too?

Regardless, that can be another issue. Merging.

ronichoudhury:
I think they are used elsewhere, which is why I didn't go ahead and delete them. Of course, those usages will likely not work on Firefox either, but I wanted to get this issue solved as quickly as I could.

brianhelba:
@ronichoudhury No problem, thanks.

PR #147, opened 2016-08-31
Eliminate the page size control

### LINKED ISSUE #133: Image gallery: Remove page size
The Page Size button isn't working properly (race conditions with paging), and may be confusing to users, so just remove it.

ronichoudhury:
Set page size to 50 statically. Later we can use a dropdown menu to offer a few choices of page size.

Fixes #133


brianhelba:
LGTM.

Issue #138, opened 2016-08-31
Image gallery: double clicking image should launch a lightbox


brianhelba:

brianhelba:
Alternatively, a "Show fullscreen" button could be added to the image detail pane, and the logic for double clicking images (which isn't really a web-native action) could be removed.

PR #149, opened 2016-08-31
Open image in new tab when double clicked

### LINKED ISSUE #137: Image gallery: double clicking image should open image in a new tab

ronichoudhury:
This is done using a bit of hackery.

When a `dblclick` event is fired, it comes with two `click` events. To get around this, I've converted the click handler to a more complicated one that receives a click, waits a set interval for a second click, then dispatches the single-click behavior (toggle selected image) if there wasn't a second click, or the double-click behavior (open image in new tab) if there was.

The one pitfall of this approach is that you can't dispatch the single-click behavior until the detection interval has elapsed. I don't believe there's any way around this. If this is unacceptable, then we'll have to find an alternate input behavior besides double-click on which to display the full size image.

Fixes #137

### DIFF
--- a/web_external/js/views/body/ImagesView/ImageWall.js
+++ b/web_external/js/views/body/ImagesView/ImageWall.js
@@ -257,6 +257,43 @@ isic.views.ImagesViewSubViews.ImageWall = Backbone.View.extend({
             .data(imageList, function (d) {
                 return d;
             });
+
+        // This click callback has to be defined here with a name because it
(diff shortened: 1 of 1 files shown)


brianhelba:
Indeed, double click isn't really a web-native action, so maybe we should move away from it eventually.

We could add a "Show fullscreen" button to the image detail pane, as part of #138 . Until then, this should be fine.

PR #158, opened 2016-09-02
Lesion Images code cleanup

msmolens:
Some miscellaneous cleanup of the lesion images views to be more consistent with the rest of the app.

More could be done:
- continue replacing code like `var self = this`
- replace usage of `Object.keys`, `forEach`, `hasOwnProperty`, etc. with Underscore.js alternatives
- avoid directly assigning properties on `window`
- use more Jade templates


brianhelba:
LGTM. I added one more commit to rename "study" to "dataset" in some places.

@msmolens Can you open a new issue to list out these other cleanup tasks / best practices that we need to apply to the remainder of the ImagesView code?

PR #162, opened 2016-09-06
Run development server at isic-archive.devel

### LINKED ISSUE #81: .dev TLD causes error in Chrome
Using `.dev` as the TLD for the hostname in the Vagrantfile causes conflicts now that `.dev` is a real gTLD (generic top-level domain): https://www.iana.org/domains/root/db/dev.html. Accessing `isic-archive.dev` in Chrome gives an `ERR_ICANN_NAME_COLLISION` error.

Renaming the hostname to `isic-archive.devel` would resolve the issue, or perhaps there is another best practice.

brianhelba:
Fixes #81.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -4,7 +4,7 @@ Vagrant.require_version ">=1.8.0"
 Vagrant.configure("2") do |config|
   config.vm.box = "ubuntu/trusty64"
 
-  config.vm.hostname = "isic-archive.dev"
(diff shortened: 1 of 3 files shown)


brianhelba:
This also renames the Vagrant-created VM.

@msmolens You should be able to rename this manually, but try this branch out before I merge it and let me know if you have any issues.

msmolens:
@brianhelba I made these same changes locally before provisioning, so feel free to merge.

PR #171, opened 2016-09-07
Web client build cleanup

brianhelba:
* Remove most customization of Girder admin UI
* Remove symlink to bower_components in web_client/extra
* Update readme to use a simpler command for "grunt watch"


brianhelba:
@msmolens This should solve your issue with Windows and symlinks inside VirtualBox.

msmolens:
@brianhelba Indeed, I can now build and run the annotation tasks in my dev environment. LGTM.

Issue #163, opened 2016-09-06
Image gallery: miscellaneous code clean up tasks


msmolens:
Some specific tasks could include:
- [x] continue replacing code like `var self = this`
- [ ] replace usage of Object.keys, forEach, hasOwnProperty, etc. with Underscore.js alternatives
- [ ] avoid directly assigning properties on `window`
- [ ] use more Jade templates, split up `imagesPage.jade`?

PR #175, opened 2016-09-12
Use _.bind(..., this) in place of var self = this

ronichoudhury:
@msmolens PTAL.

Addresses #163


ronichoudhury:
I should probably not have touched `dermapp-viewer.js`, as it's in the `static` directory. Let me know if I should revert that commit.

brianhelba:
Yeah, please back out the changes to `dermapp-viewer.js`. That's legacy code that will go away when we get GeoJS online, but I want to avoid any risks of breaking it until then

ronichoudhury:
I force pushed to drop that commit, so update your checkouts accordingly.

brianhelba:
LGTM.

PR #176, opened 2016-09-13
Use underscorejs functions in codebase

ronichoudhury:
Depends on #175 

Addresses #163


ronichoudhury:
@msmolens what other constructs should I underscorify?

PR #178, opened 2016-09-13
Use a lightbox to display image detail

### LINKED ISSUE #138: Image gallery: double clicking image should launch a lightbox

ronichoudhury:
This PR removes the double click logic that used to open an image in a new tab. In its place, the detail menu that appears when an image is selected includes two new buttons: one labeled "fullscreen" and another labeled "open in new window".

The "open in new window" button takes over the behavior of the double click action - a new window is opened to display the image in full resolution.

The "fullscreen" button opens a Bootstrap modal without a header or footer (so it looks just like what a dedicated lightbox library would provide) showing the full resolution image. The modal is sized appropriately to accommodate the image, and it appears centered in the browser window, even on resizing.

Closes #138 

Depends on #177

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -126,8 +126,6 @@ module.exports = function (grunt) {
                 'jquery/dist/jquery.min.js',
                 'flatstrap/dist/js/bootstrap.min.js',
                 'flatstrap/dist/css/bootstrap.min.css',
-                'flatstrap/dist/fonts/glyphicons-halflings-regular.woff',
-                'flatstrap/dist/fonts/glyphicons-halflings-regular.ttf',
                 'font-awesome/css/font-awesome.min.css',
                 'angular/angular.min.js',
                 'angular-resource/angular-resource.min.js',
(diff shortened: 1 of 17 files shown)


brianhelba:
Can you make it so left-clicking inside the modal also closes it? We may want to remove this behavior in the future though, so please make the event bindings separate. 

Also, the buttons are lacking margins, don't fit the color scheme, and take up more vertical space than necessary. Could we make the buttons smaller (you can add tooltips for longer descriptions) and move them onto the same line as the title? I think the lightbox button is more important than the pop-out one, so maybe "fullscreen" could have text and the other one just has an icon and tooltip. The color might also be nicer with white-background buttons. 

I've made this branch live, as it's a solid improvement, but I'll wait to merge to master until these final changes are in.

ronichoudhury:
@brianhelba I've tried to address all your requests. I think it looks really nice now.

One decision I made was to do icon/tooltip for both buttons. I don't think mixing icons and text would look good here.

Let me know what you think.

brianhelba:
This looks great now. Merging.

Note, the styling on the various buttons in the image gallery doesn't totally match (different border style, background color on hover, presence of tooltips), but I'm sure this will unify as we continue to refactor and pay more attention to style.

ronichoudhury:
On Fri, Sep 16, 2016 at 4:40 PM Brian Helba <notifications@github.com>
wrote:

> Note, the styling on the various buttons in the image gallery doesn't
> totally match (different border style, background color on hover, presence
> of tooltips), but I'm sure this will unify as we continue to refactor and
> pay more attention to style.
>
If you file an issue about this, then it will be sure not to fall off the
radar.

PR #189, opened 2016-09-21
Add a base script for access via the HTTP API

brianhelba:


brianhelba:
@aadikalloo PTAL. You may find this useful.

PR #198, opened 2016-09-23
Fix icons in Phase 1

brianhelba:


ronichoudhury:
@brianhelba if you could tag me (or ask for review) on PRs related to my tasks, that would be helpful (don't want things to fall off my radar unnecessarily).

Issue #135, opened 2016-08-31
Image gallery: Image layout needs improvement


brianhelba:
* Opening and closing the image detail pane causes large gaps to appear in the layout
* The images aren't appearing in the same sorted order that they're received from the server.

brianhelba:
Fixed.

PR #200, opened 2016-09-26
Use img elements to hold thumbnails

ronichoudhury:
This ditches a lot of the logic for emplacing SVG image elements and catching AJAX calls to fill in their image contents, opting instead for ordinary `img` elements, allowing the browser to place them where it wants to.

Addresses #135


ronichoudhury:
The major issue with this change is that a selected image now reflows instantaneously to its proper place "downstream" when the details pane opens up (click on the first image in the second row to trigger the behavior).

I figured we could address that problem in a separate PR if this one is good, but I'll defer to @brianhelba.

PR #223, opened 2016-10-07
Fix segmentation dropdown

### LINKED ISSUE #215: Segmentations dropdown is not working
The API call to `/segmentation` appears to be made with some incorrect query string parameters.

brianhelba:
Fixes #215

Also unify the default sort parameter between the JS collection and the server's API for Segmentation.


brianhelba:
FYI @msmolens

PR #238, opened 2016-10-21
Disable user selection in image wall

### LINKED ISSUE #233: Prevent mouse selection of gallery thumbnails
It should not be possible to click and drag to select gallery thumbnail elements, as the selection color makes the images difficult to see.

Alternatively, the color change on selection could be disabled.

ronichoudhury:
Closes #233


brianhelba:
LGTM.

PR #240, opened 2016-10-22
Add GeoJS viewer to gallery

### LINKED ISSUE #180: Image gallery: make lightbox use GeoJS for zoomable images

brianhelba:
Fixes #180.

TODO:
- [x] Fix the sizing of `ImageFullscreenWidget` modals


brianhelba:
@msmolens While working on this, I noticed that [`studyResultsPage.jade`](https://github.com/ImageMarkup/isic-archive/blob/b1970c25795c3e1fb21840376b83ea0c5bc38d7c/web_external/templates/body/studyResultsPage.jade#L45) seems to contain two image containers divs (though only one is actually used). Is this cruft?

brianhelba:
@ronichoudhury PTAL at changes in https://github.com/ImageMarkup/isic-archive/pull/240/commits/df4a2c1008e35337412d2f8da03af57ddadfee5d

brianhelba:
@ronichoudhury Also PTAL at https://github.com/ImageMarkup/isic-archive/pull/240/commits/c98a308d6991a5381df0d4f50c066634543debcf

msmolens:
@brianhelba I see how the image containers in `studyResultsPage.jade` can be confusing--there are 3 comments that say "Image container". But all 3 are still used. The containers at the bottom of the file (for the large images) are tied to tab selection. They aren't nested under `.tab-content` because sizing/layout wasn't working as well when I tried that.

web_external/js/views/body/ImagesView/ImageWall.js
```diff
@@ -25,52 +21,49 @@ isic.views.ImagesViewSubViews.ImageWall = isic.View.extend({
...
             .on('click', _.bind(function (d) {
                 this.clearTooltips();
                 if (d3.event.shiftKey) {
-                    var image = new isic.models.ImageModel({
-                        _id: d
-                    });
                     new isic.views.ImageFullscreenWidget({ // eslint-disable-line no-new
                         el: $('#g-dialog-container'),
-                        model: image,
+                        model: this.model.images.findWhere({_id: d.id}),
```
ronichoudhury: Why not just stuff the model into the data binding in the hash in line 31ff. above? That would eliminate this quadratic time lookup approach.

ronichoudhury:
c98a308 looks ok.

web_external/js/views/body/ImagesView/ImageWall.js
brianhelba: I didn't take the time to fully test whether the Backbone models would cleanly be stored in D3 `data` (whether it's a shallow copy, if model/collections updates get applied in the correct order, etc.).

Assuming it works, storing the model directly would be an improvement. However, as-is, it's just a `O(n)` lookup each time a modal dialog is opened (the tooltips are now constant time, which might have solved some bugs with them).

ronichoudhury: Ok. LGTM.

PR #244, opened 2016-10-24
Show featureset version numbers in headings

### LINKED ISSUE #229: Add version numbers to featureset page tab names
Since there may be multiple versions of identically-named featuresets, the collapse headers on the featureset page should include the version value.

msmolens:
Display featureset version number in heading to distinguish featuresets that have the same name.

Fixes #229


brianhelba:
LGTM.

PR #245, opened 2016-10-24
Clarify multi-component names on featureset page

### LINKED ISSUE #230: Clarify multi-component names on featureset page
Right now, name components are just being joined with a comma for display. It should also be possible for a user to easily tell what the separate components of a name are.

msmolens:
Add a separator between feature name components. The separator suggests a hierarchical relationship between the components.

Fixes #230

### DIFF
--- a/web_external/stylesheets/body/featuresetPage.styl
+++ b/web_external/stylesheets/body/featuresetPage.styl
@@ -17,6 +17,11 @@
       padding-top 10px
       font-size 16px
 
+      .isic-featureset-table-feature-name-component
+        padding 2px 6px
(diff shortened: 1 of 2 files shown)


brianhelba:
I like the use of the `>` to delimit name components, as I think it's unlikely that symbol will be natively used in feature names. It's also nice that it's rendered as an icon, so it isn't selectable as normal text.

However, can we add a little more styling (font settings, background shading, etc.) to distinguish the textual name components from the delimiter?

msmolens:
@brianhelba I added some background color to better differentiate the name components.

PR #248, opened 2016-10-25
Add version numbers to featureset search

### LINKED ISSUE #246: Add version numbers to featureset search dialog

msmolens:
Fixes #246

### DIFF
--- a/web_external/js/views/body/CreateStudyView.js
+++ b/web_external/js/views/body/CreateStudyView.js
@@ -39,8 +39,6 @@ isic.views.CreateStudyView = isic.View.extend({
             parentView: this
         }).on('g:resultClicked', this._addUser, this);
 
-        var formatFeatureset = this.formatFeatureset;
-
         this.featuresetSearchWidget = new girder.views.SearchFieldWidget({
             placeholder: 'Search featuresets...',
             modes: ['prefix', 'text'],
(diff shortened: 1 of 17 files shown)


brianhelba:
One thing I realized is that `girder.Model` has [a `name()` function](https://github.com/girder/girder/blob/ae461ee437b9a788cd2ed2d2e47b567ac0996de8/clients/web/src/model.js#L28), which defaults to `this.get('name');`, but could be overridden by subclasses. Perhaps we should just be more consistent about only using `Model.name` to get rendered names, then override that function in `isic.models.FeaturesetModel`.

@msmolens What do you think?

msmolens:
@brianhelba I think at least for `FeaturesetModel` it makes sense to override `name()`. I will update this PR accordingly.

brianhelba:
@msmolens Per our discussion, I'll work on this.

brianhelba:
@msmolens PTAL. This goes a little beyond the scope of the original issue, but I think it will be more maintainable to use `.name()` to render our JS generally.

web_external/js/views/body/CreateStudyView.js
```diff
@@ -46,7 +46,7 @@ isic.views.CreateStudyView = isic.View.extend({
             getInfoCallback: function (type, obj) {
                 if (type === 'featureset.isic_archive') {
                     return {
-                        text: obj.name,
+                        text: (new isic.models.FeaturesetModel(obj)).name(),
```
msmolens: I'd prefer to assign the new object to a named variable to simplify the syntax, but I'm ok if we don't change it too.

msmolens:
@brianhelba Nice, much cleaner! Please take a look at the commits I added.

PR #255, opened 2016-10-26
Add interactive viewer to the study results page

### LINKED ISSUE #250: Add interactive viewer to the study results page
Replace `StudyResultsImageView` with use of `ImageViewerWidget` on the study results page.

msmolens:
Fixes #250

### DIFF
--- a/web_external/js/views/body/StudyResultsView.js
+++ b/web_external/js/views/body/StudyResultsView.js
@@ -347,6 +347,14 @@ isic.views.StudyResultsFeatureImageView = isic.View.extend({
         this.listenTo(this.model, 'change', this.render);
     },
 
+    setVisible: function (visible) {
+        if (visible) {
+            this.$el.removeClass('hidden');
+        } else {
+            this.$el.addClass('hidden');
+        }
+    },
+
     render: function () {
         var featureId = this.model.get('featureId');
         var annotationId = this.model.get('annotationId');
@@ -440,6 +448,12 @@ isic.views.StudyResultsLocalFeaturesView = isic.View.extend({
 
 // View for an image
 isic.views.StudyResultsImageView = isic.View.extend({
+    setVisible: function (visible) {
+        if (visible) {
+            this.$el.removeClass('hidden');
+        } else {
+            this.$el.addClass('hidden');
+        }
     initialize: function (settings) {
         this.listenTo(this.model, 'change', this.render);
     },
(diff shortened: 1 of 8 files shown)


web_external/js/views/widgets/ImageViewerWidget.js
```diff
@@ -10,16 +22,40 @@ isic.views.ImageViewerWidget = isic.View.extend({
             this.tileHeight = resp.tileHeight;
             this.sizeX = resp.sizeX;
             this.sizeY = resp.sizeY;
+            this.destroyViewer();
```
brianhelba: Don't we want to call `this.destroyViewer();` before setting the tile metadata here, since `destroyViewer` is going to `delete` all the tile metadata if a viewer exists?

web_external/js/views/widgets/ImageViewerWidget.js
```diff
@@ -10,16 +22,40 @@ isic.views.ImageViewerWidget = isic.View.extend({
...
     },
 
     render: function () {
+        // Do nothing if model is not set
+        if (!this.model.id) {
+            return this;
+        }
+
+        // Ensure tile info is available before rendering
+        if (_.isUndefined(this.sizeX) || _.isUndefined(this.sizeY)) {
```
brianhelba: You can just test a single property. Either all will be set or none will.

web_external/js/views/widgets/ImageViewerWidget.js
```diff
@@ -1,6 +1,18 @@
 isic.views.ImageViewerWidget = isic.View.extend({
     initialize: function (settings) {
         /* "model" and "el" must be passed as settings */
+
+        this.renderedModelId = null;
+
+        this.listenTo(this.model, 'change', this.fetchTileInfo);
```
brianhelba: Note that if you instantiate a `ImageViewerWidget` with an `el` and `model`, then the model gets changed, the `ImageViewerWidget` instance will spontaneously render itself, even if `render()` was not yet explicitly called by the owner. I don't think there's an easy way around this though, and it's probably not going to be an issue, given the ways we're using `ImageViewerWidget` right now.

web_external/js/views/body/StudyResultsView.js
```diff
@@ -436,16 +448,29 @@ isic.views.StudyResultsLocalFeaturesView = isic.View.extend({
 
 // View for an image
 isic.views.StudyResultsImageView = isic.View.extend({
-    initialize: function (settings) {
-        this.listenTo(this.model, 'change', this.render);
+    setVisible: function (visible) {
+        if (visible) {
+            this.$el.removeClass('hidden');
+
+            if (this.imageViewerWidget) {
```
brianhelba: When would we be calling `.setVisible(true)` on a `StudyResultsImageView` that hasn't yet be rendered? Or is this just to deal with a race condition during initialization?

web_external/js/views/widgets/ImageViewerWidget.js
msmolens: Thanks for catching this.

msmolens:
@brianhelba Thanks for the review. Please review the latest changed.

PR #260, opened 2016-10-28
Add owner field to Upload Dataset

### LINKED ISSUE #168: Upload page: Add a field for dataset owner's name
There will now be 3 name-related fields:
- Owner's name (this new text field)
- Uploader's name (implicitly from the logged-in user)
- Attribution name (optional, could be "Anonymous")

msmolens:
Fixes #168

### DIFF
--- a/server/api/dataset.py
+++ b/server/api/dataset.py
@@ -92,6 +92,7 @@ class DatasetResource(Resource):
         .param('uploadFolderId', 'The ID of the folder that contains images '
                'and metadata.')
         .param('name', 'Name of the dataset.')
+        .param('owner', 'Owner of the dataset.')
         .param('description', 'Description of the dataset.', required=False,
                paramType='form')
         .param('license', 'License of the dataset.', required=False,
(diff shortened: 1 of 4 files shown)


server/api/dataset.py
```diff
@@ -128,6 +129,10 @@ def ingestDataset(self, params):
                 'Invalid upload folder ID.', 'uploadFolderId')
 
         name = params['name'].strip()
+        owner = params['owner'].strip()
+        if not owner:
```
brianhelba: Why aren't we using `self.requireParams` here (and below)? Is it because we want to `strip` the string first, or because we want a more descriptive error message?

.agignore
```diff
@@ -0,0 +1 @@
+custom
```
brianhelba: What's this file?

server/api/dataset.py
msmolens: Since I made `owner` a required parameter, it was an oversight that I didn't include it in the initial call to `requireParams`. Nevertheless, it's explicitly checked to be nonempty again after `strip`, since `requireParams` doesn't inspect the content.

.agignore
msmolens: It doesn't belong. I took measures to make sure this doesn't happen in the future.

msmolens:
When we get to changing the layout, we could consider a wizard style instead of displaying all fields simultaneously.

PR #267, opened 2016-11-02
Fix "sticky tooltips" problem

### LINKED ISSUE #251: Tooltips get stuck on image gallery
On the image gallery image thumbnail tooltips can get stuck. This happens intermittently for me, but seems to be triggered easily by rapidly clicking images.

Browser: Chrome, Firefox

image

ronichoudhury:
Closes #251 

Also fixes possibly related (before now, unobseved) problem of "flickering" in details pane when images were rapidly selected/deselected.

### DIFF
--- a/web_external/js/views/body/ImagesView/ImageDetailsPane.js
+++ b/web_external/js/views/body/ImagesView/ImageDetailsPane.js
@@ -11,7 +11,7 @@ isic.views.ImagesViewSubViews.ImageDetailsPane = isic.View.extend({
     initialize: function (settings) {
         this.image = settings.image;
 
-        this.listenTo(this.image, 'changed:_id g:fetched g:error', this.render);
+        this.listenTo(this.image, 'g:fetched g:error', this.render);
 
         this.segmentationsDisplayView = new isic.views.SegmentationsDisplayView({
             image: this.image,
--- a/web_external/js/views/body/ImagesView/ImageWall.js
+++ b/web_external/js/views/body/ImagesView/ImageWall.js
@@ -14,6 +14,9 @@ isic.views.ImagesViewSubViews.ImageWall = isic.View.extend({
     render: _.debounce(function () {
         var self = this;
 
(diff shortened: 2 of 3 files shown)


ronichoudhury:
@msmolens PTAL

msmolens:
image

brianhelba:
@msmolens The theory behind the first commits by @ronichoudhury was that slowness in rendering tooltips was causing some sort of race condition where they aren't getting cleaned up.

However, I was also able to reproduce the tooltip problem just by waving and clicking my mouse wildly (but not changing the page) with this branch applied.

So, I've added a final commit that just manually removes the tooltip-containing element from the DOM when tooltips are explicitly cleared (now on click or page switch). This will not prevent tooltips from sticking as a result of wild mouse waving, but does provide a mechanism for such stray tooltips to be cleaned up on the next explicit clear. I'm also concerned by the fact that [the documentation for `.tooltip('hide')`](http://getbootstrap.com/javascript/#tooltip-hide) states: `Hides an element's tooltip. Returns to the caller before the tooltip has actually been hidden`, which means we could be manually deleting the tooltips before Bootstrap has a chance to complete a pending hide operation (which might have succeeded).

@ronichoudhury @msmolens PTAL and let me know if this final commit resolves the issue or causes you to see any console errors.

msmolens:
@brianhelba The latest commit looks like it fixes the problem. LGTM.

FYI I also tried setting `trigger: 'hover'` like in https://github.com/ImageMarkup/isic-archive/blob/3301e004b7f0b761bd61537192e3ff53542efda0/web_external/js/views/layout/HeaderView.js#L30-L31, as opposed to the default `trigger: 'hover focus'` but that didn't solve any of the sticky tooltip problems on its own.

brianhelba:
@ronichoudhury @msmolens I made one more change to set `trigger: 'hover'` on tooltips, just as a defensive measure. We don't really expect images to be getting focused anyway.

I'll merge in a bit, once the tests pass.

PR #272, opened 2016-11-03
Add Terms Of Use to upload page

### LINKED ISSUE #271: Add real TOU to "Contributor license agreement" of upload page
This should use templating if possible, to avoid copy and pasting of the TOU text.

msmolens:
Tasks:
- [x] Add TOU as Contributor license agreement
- [x] Support Download as text
- [x] Support Download as PDF

Fixes #271


web_external/extra/documents/Terms of Use.txt
```diff
@@ -0,0 +1,124 @@
+International Skin Imaging Collaboration (�ISIC�)
```
brianhelba: I think the curly quotes in the original document aren't being translated to straight quotes here; this also seems to be the case on the `legal/terms_of_use.html` page.

I think we should fix it here, just because it will be more apparent to users who download this and open it in notepad. The `legal/terms_of_use.html` can wait until we do a proper style pass.

web_external/extra/documents/Terms of Use.txt
```diff
@@ -0,0 +1,124 @@
...
+
+10. Third Party Links. The Site may contain links to other web sites that are operated by third parties and that are identifiable by a User of the Site as being operated by third parties ("Links"). You acknowledge that MSKCC does not endorse the Links and is not responsible for any content of the Links. You also acknowledge that the respective owners of the Links neither endorse nor are affiliated with MSKCC.  
+
+11. Site Privacy Policy. 
+
+11.1. By using the Site, you are agreeing to the terms set forth in this Site Privacy Policy, which may be found at the following link: https://isic-archive.com/privacy_policy .
```
brianhelba: This link needs to be fixed.

brianhelba:
The TOU PDF is 572KB in size. I'm hesitant to commit binary blobs this large to the repo, especially ones that will be iterated upon with several versions. Is there anything that can be done to make the PDF smaller?

Also, the URL for the privacy policy in the PDF is also incorrect.

msmolens:
@brianhelba The text document now uses straight quotes and is wrapped at 80 characters. The PDF is smaller, but still large at 161KB. Privacy Policy links are fixed.

msmolens:
Another option is to store the PDF as a Girder item and and add option in the plugin's configuration. Obviously that's more general, but could complicate deployment.

brianhelba:
@msmolens I'd like to make sure that we get the content of the PDF exactly right before merging it to master, to prevent bloating our repo size with multiple revisions of a binary blob. This is not a publicly facing page, so I think we can take a little longer to revise and get this right.

So, let's hold on merging this branch while we gather feedback on the legal text in general. We can always temporarily merge it on the live site only, for demonstration purposes.

brianhelba:
@msmolens I've emailed you some new contact info. Please update everything with that, then we should be able to merge.

msmolens:
@brianhelba I updated the contact information and updated the commit with the PDF.

msmolens:
Rebased to build on top of recent Travis CI fix.

msmolens:
Latest version changed the text version of TOU to ASCII encoding instead of ISO8859-1. "§" is replaced by "S".

PR #274, opened 2016-11-03
Use ISIC user permissions to determine access

brianhelba:


brianhelba:
@msmolens PTAL.

web_external/js/app.js
```diff
@@ -1,5 +1,59 @@
...
+        // before "g:login", but "g:login.success" isn't triggered after a new
+        // user registers (which is probably a bug)
+
+        // "currentUser" must be updated before all rendering, so unbind the
+        // default binding of "g:login" to "this.login", and ensure that
+        // updating "currentUser" is the event bound to "g:login"; however,
+        // it's an undocumented property of Backbone that events get called in
+        // the order they were bound, so this isn't a great solution
+        girder.events.off('g:login', this.login);
+        girder.events.on('g:login', this._updateCurrentUser(), this);
```
msmolens: This event isn't attached correctly--it's calling `updateCurrentUser()`. Because of this I see console errors if I log out and log in.

web_external/js/app.js
```diff
@@ -1,5 +1,59 @@
...
+        // "g:login-changed" is triggered when users change passwords
+        girder.events.on('g:login-changed', this._updateCurrentUser(), this);
```
msmolens: Same here.

brianhelba:
@msmolens Fixed; I think my IDE did that to me as I was refactoring.

PR #276, opened 2016-11-08
Make the Featureset REST endpoint also sort by version

brianhelba:


brianhelba:
@msmolens PTAL.

server/api/featureset.py
```diff
@@ -47,6 +48,14 @@ def find(self, params):
         # TODO: make the default sort lowerName (after adding that field)
         limit, offset, sort = self.getPagingParameters(params, 'name')
 
+        # Since names may be identical, add an additional sort field for
+        # convenience, especially as the REST interface cannot specify multiple
+        # fields
```
msmolens: > especially as the REST interface cannot specify multiple fields

Yes, that's unfortunate.

PR #281, opened 2016-11-08
Replace formatUser with UserModel.name()

brianhelba:


brianhelba:
@msmolens PTAL.

At some point, we should start caching the models created by owner model accessor methods, but I don't see any performance issues at the moment.

web_external/js/views/body/StudyResultsView.js
```diff
@@ -589,10 +587,7 @@ isic.views.StudyResultsView = isic.View.extend({
             this.images.reset(imageModels);
 
             // Populate users collection
-            var userModels = _.map(this.study.get('users'), function (user) {
-                return new girder.models.UserModel(user);
-            });
-            this.users.reset(userModels);
+            this.users.reset(this.study.users());
```
msmolens: `reset()` can't handle a `Collection`, but we could use `this.study.users().models`.

web_external/templates/body/studyPage.jade
```diff
@@ -10,8 +10,8 @@
   .isic-listing-content.isic-study-users-content
     if study.get('users').length
       ul
-        each user in study.get('users')
-          li= formatUser(user)
+        each user in study.users().toArray()
```
msmolens: `study.users().models` gives raw access to the collection's model array, which would be more efficient.

brianhelba:
@msmolens Fixed both places.

PR #283, opened 2016-11-08
Fix Travis failures

brianhelba:


brianhelba:
Fixes the problem. Merging.

PR #284, opened 2016-11-08
Hide/show histograms

ronichoudhury:


brianhelba:
@ronichoudhury PTAL, I think this should fix the tooltip issue. It's not the best solution (as image tooltips also get cleared when the histogram is re-rendered), but more specific fix will require even more complex logic or a significant amount of additional refactoring. I'd rather just wait for the inevitable refactor, which will improve the situation in lots of ways.

@msmolens PTAL at the functionality. You're welcome to peek at the code too, but we have a lot more planned refactoring to go.

web_external/js/views/body/ImagesView/HistogramPane.js
```diff
@@ -71,6 +71,8 @@ isic.views.ImagesViewSubViews.HistogramPane = isic.View.extend({
             });
         attributeSections.select('.content').each(_.bind(function (d) {
             var histogramId = isic.shims.makeValidId(d + '_histogramContent');
+            // Destroy all tooltips before re-rendering, as a precaution
+            $('.tooltip').remove();
```
ronichoudhury: If you wanted to leave the image tooltips alone, you could just class them differently and scope this selector to just the histogram tooltips right?

msmolens:
The behavior seems wrong when I toggle a checkmark on a histogram bar. It looks like the entire view re-renders, histogram collapsed states are lost, and the view scrolls to the top. Is that a known behavior pending future refactoring?

ronichoudhury:
@brianhelba this is a good illustration of why I'm against the "tear-down-and-rebuild" notion of `render()`. I might be mistaken about the connection between my position and Max's observation, but asking the histogram to re-render when a filter state changes is either (1) the wrong thing to do (instead of re-rendering, just call a function that asks for and renders new histogram bars) or (2) perfectly reasonable, in which case `render()` should not be a destructive rebuild of the whole view.

brianhelba:
@ronichoudhury @msmolens Indeed, there are several different ways to fix this. Regardless of how we proceed, the old gallery code had major issues, so I can chalk this up to growing pains of refactoring.

I'd prefer to see the gallery code (like code in other parts of the archive):
* Render (or "tear-down-and-rebuild") everything when `render()` is called
* Expose other methods (e.g. `updateBars` that tweak specific parts of an already-rendered view)
* Hook up events emitted by model changes to the specific-update methods
The problem we're seeing now is that any changes to the `overviewHistogram` and `filteredSetHistogram` are causing the entire panes (and all histograms within) to re-`render()`. I'll try to fix this ASAP, and roll back the live site to an earlier master if necessary.

As I see it, the other approach is to put a lot more logic into `render()`, to determine what already exists, what has changed in models, and how the DOM needs to be updated accordingly. Ultimately, this is probably going to be a similar number of LOC, but keeping it split into separate functions makes it more maintainable.

Of course, I'm open to alternative approaches, but for as long as we're within Backbone, I personally don't know of anything better.

PR #285, opened 2016-11-09
Reorganize all external web app files

brianhelba:
This is going to have merge conflicts with almost all other WIP-branches, and it's not building correctly yet. Still, I think this will ultimately make things more maintainable, as the app continues to grow.


brianhelba:
@msmolens PTAL.
@ronichoudhury PTAL.

brianhelba:
The other thing to remember is that Jade files are still loaded by their non-qualified filename, so those can't be renamed. 

Once we move to Webpack, I'm open to reconsideration of the file names though.

ronichoudhury:
Sounds good.

PR #286, opened 2016-11-10
Refactor histograms into FacetViews

brianhelba:


brianhelba:
@ronichoudhury PTAL. We can take a look at this together if you'd like an outline of the changes.

web_external/js/views/body/ImagesView/ImagesFacetsPane.js
```diff
@@ -0,0 +1,48 @@
...
+            }
+
+            var facetView = new FacetView({
+                model: this.model,
+                facetName: facetName,
+                parentView: this
+            });
+
+            this.facetViews[facetName] = facetView;
+            this.$el.append(facetView.el);
```
brianhelba: Hey @msmolens , I found it useful here to let Backbone create the `el` for each `FacetView` from scratch, then append that new element into the existing main DOM. I have the feeling that we should probably take better advantage of this pattern in other parts of the code.

msmolens: For dynamic, repeated elements we typically use iteration in the jade templates and avoid explicit code like `this.$el.append()`, but I agree we should keep this style in mind.

web_external/js/views/body/ImagesView/ImagesFacetsPane.js
```diff
@@ -0,0 +1,48 @@
+isic.views.ImagesFacetsPane = isic.View.extend({
+    render: function () {
+        if (!this.addedCollapseImage) {
```
ronichoudhury: I agree. We kept it around in this PR just out of caution, and resolved to cut out unneeded assets and associated code in a separate sweep.

PR #291, opened 2016-11-10
Add links to Terms of Use and Privacy Policy in footer

### LINKED ISSUE #279: Add links to TOU and Privacy Policy to site footer
Perhaps all 3 policy links should be moved to the center of the footer?

msmolens:
Fixes #279 

One thing to note is that the footer doesn't currently collapse very reasonably when the available width is small. I investigated using Bootstrap's grid system, but AFAIK that doesn't allow full-width containers out of the box, so the left and right text wouldn't be aligned to the left and right of the window.


brianhelba:
Doesn't `.container-fluid` give you a full-width container?

PR #295, opened 2016-11-11
Display the "diagnosis_confirm_type" facet

brianhelba:


brianhelba:
Currently, this results in histogram labels that are too long. We should add some new logic to clip overly-long histogram labels (using something like `text-overflow: ellipsis;`) then display them in a tooltip when moused over.

ronichoudhury:
@brianhelba PTAL

ronichoudhury:
attn: @brianhelba - I force pushed some fixes, including one for the failing style test. Just be careful when you go to update your branch.

PR #298, opened 2016-11-15
Rename special histogram bin values

brianhelba:
This renames the special bin values  `"__undefined__"`, `"__null__"`, `"__NaN__"`, `"__other__"` to reduce the likelihood of collisions with actual data bin values.

This also defaults to [not putting excess histogram bins into `"__other__"`](https://github.com/ImageMarkup/isic-archive/pull/298/files#diff-7a5c68fa12d300ef580a972414f7ddf0R83), when `numBins` is not defined for a histogram. Previously, any additional values, after the [first 10 distinct ones](https://github.com/ImageMarkup/isic-archive/pull/298/files#diff-ae0e0f9d696bd994644cb93b0cab2c20R174), would be placed into an `"other"` bin.

This ensures that [`"__null__"` values sent by the client](https://github.com/ImageMarkup/isic-archive/pull/298/files#diff-1bd33893522244904a198aa9d73f68edR163) when creating a filter are converted into proper `None` values, though this does not have any effect with the current client configuration (which transforms all `"__null__"` into `"__undefined__"` or `"__NaN__"`).

This fixes a minor bug where the client was [transforming `"__null__"` into](https://github.com/ImageMarkup/isic-archive/pull/298/files#diff-5ab9722391177e1811dc3aca6cbbdd3dL110) the wrong one of `"__undefined__"` or `"__NaN__"`.

### DIFF
--- a/server/histogram_utility/__init__.py
+++ b/server/histogram_utility/__init__.py
@@ -171,7 +171,7 @@ class HistogramUtility(object):
             binSettings[attrName]['specialBins'] = specialBins
 
             # Get user-defined or default number of bins
-            numBins = binSettings[attrName].get('numBins', 10)
+            numBins = binSettings[attrName].get('numBins', 0)
             binSettings[attrName]['numBins'] = numBins
 
             # For ordinal binning, we need some more details:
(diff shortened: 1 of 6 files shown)


brianhelba:
@ronichoudhury PTAL.

brianhelba:
@ronichoudhury I've updated the description with a more detailed list of changes.

ronichoudhury:
The code all looks good to me but I'm not sure what I can do to "test" it live.

brianhelba:
I've already tested it on live.

PR #299, opened 2016-11-15
Add a page to request Dataset Contributor access

### LINKED ISSUE #287: Add a page to request Dataset Contributor access
Users without the `createDataset` permission should still be able to see the "Upload Dataset" button in the menu. When clicked, such users will be shown a page with something like: 
```Thank you for your interest in contributing to the ISIC Archive. We accept X type of dermoscopic images, with Y metadata, under CC-0, CC-BY- and CC-BY-NC license. If you have images to contribute, please click this button.```
Clicking the button will make a request for the user to join the "Dataset Contributors" group.

Upon any requests to join the "Dataset Contributors" group, an email should be sent to either the site admins or the moderators / administrators of the group (or perhaps both), as Girder's notification for pending group requests is not very obvious.

### PR DESCRIPTION
**Pull request by msmolens**
<strike>WIP because of placeholder text in `uploadDatasetRequest.jade`. What should the final text be?</strike>

Fixes #287

### DIFF
--- a/server/api/user.py
+++ b/server/api/user.py
@@ -102,7 +102,7 @@ def requestCreateDatasetPermission(params):
                 })
             mail_utils.sendEmail(
                 to=emails,
-                subject='ISIC Archive Dataset Contributor Request',
+                subject='ISIC Archive: Dataset Contributor Request',
                 text=html)
 
     return resp
--- a/server/api/user.py
+++ b/server/api/user.py
@@ -18,6 +18,11 @@
 ###############################################################################
 
 from girder import events
+from girder.api import access
+from girder.api.describe import Description, describeRoute
+from girder.api.rest import getCurrentUser, RestException
+from girder.constants import AccessType
+from girder.utility import mail_utils
 from girder.utility.model_importer import ModelImporter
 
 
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: server/api/user.py [hunk 1]
```diff
@@ -51,10 +56,64 @@ def onPostUser(event):
...
+            html = mail_utils.renderTemplate(
+                'datasetContributorRequest.mako',
+                {
+                    'user': currentUser,
+                    'group': group,
+                    'host': host,
+                })
+            mail_utils.sendEmail(
+                to=emails,
+                subject='ISIC Archive Dataset Contributor Request',
```
brianhelba: Add a colon to this subject, to make it `'ISIC Archive: Dataset Contributor Request'`

server/api/user.py
```diff
@@ -51,10 +56,64 @@ def onPostUser(event):
...
+
+
+def attachUserApi(user):
     events.bind('rest.get.user/authentication.after',
                 'onGetUserAuthentication', onGetUserAuthentication)
     events.bind('rest.get.user/me.after',
                 'onGetUserMe', onGetUserMe)
     events.bind('rest.post.user.after',
                 'onPostUser', onPostUser)
+    user.route('POST', ('requestCreateDatasetPermission',),
```
brianhelba: This route name is a little wordy, but I can't think of a better alternative right now. I don't think this will be an important part of our documented API surface, so I'm fine with changing it later.

server/mail_templates/datasetContributorRequest.mako
```diff
@@ -0,0 +1,21 @@
+<%include file="_header.mako"/>
+
+<p>
+This is an automated message from ISIC Archive.
```
brianhelba: Make this: ...`from the ISIC Archive`...

web_external/js/views/body/UploadDatasetView.js
```diff
@@ -257,10 +255,61 @@ isic.views.UploadDatasetView = isic.View.extend({
...
+    },
+
+    requestCreateDatasetPermission: function () {
+        girder.restRequest({
+            path: 'user/requestCreateDatasetPermission',
+            type: 'POST'
+        }).done(_.bind(function (resp) {
+            if (_.has(resp, 'extra') && resp.extra === 'hasPermission') {
+                // Directly update user permissions
+                girder.currentUser.setCanCreateDataset(true);
```
brianhelba: When is this going to happen? If the user was granted permission to upload between the time that they last logged in or refreshed the front-end app and the time that they're clicking this button?

web_external/js/utilities.js
```diff
@@ -0,0 +1,42 @@
...
+ * Show an alert dialog with a single button.
+ * @param [text] The text to display.
+ * @param [buttonText] The text for the button.
+ * @param [buttonClass] Class string to apply to the button.
+ * @param [escapedHtml] If you want to render the text as HTML rather than
+ *        plain text, set this to true to acknowledge that you have escaped any
+ *        user-created data within the text to prevent XSS exploits.
+ * @param callback Callback function called when the user clicks the button.
+ */
+isic.showAlertDialog = function (params) {
```
brianhelba: Is there a reason this isn't a Backbone view, like [ImageFullscreenWidget](https://github.com/ImageMarkup/isic-archive/blob/master/web_external/js/views/widgets/ImageFullscreenWidget.js)?

web_external/js/views/body/UploadDatasetView.js
msmolens: Exactly. Perhaps it's not good, because this will only happen if they click the Request Access button again, but I wanted some solution for the user to discover they have permission short of refreshing the app or logging out. The other solution is to revert to updating the user on the client in the upload dataset route.

web_external/js/utilities.js
msmolens: This is modeled off [girder.confirm]( https://github.com/girder/girder/blob/599a9dbe7180f3f02bb7605205ec1c88f92fd368/clients/web/src/utilities/MiscFunctions.js#L108).

msmolens:
@brianhelba I addressed the changes we discussed and filed https://github.com/ImageMarkup/isic-archive/issues/302 to look at later.

PR #304, opened 2016-11-30
Allow the segmentation task listing API to return a full list of images

brianhelba:


brianhelba:
@msmolens PTAL.

brianhelba:
@msmolens Fixed.

PR #305, opened 2016-12-01
Improve Python 3 compatability

brianhelba:
* Use the next() builtin instead of the .next() method of iterators
* Use compatability methods to iterate over dicts

### DIFF
--- a/server/upload.py
+++ b/server/upload.py
@@ -180,7 +180,7 @@ def handleCsv(dataset, prereviewFolder, user, csvFile):
                     filename)
                 continue
             else:
-                imageItem = imageItems.next()
+                imageItem = next(imageItems)
 
             unstructuredMetadata = imageItem['meta']['unstructured']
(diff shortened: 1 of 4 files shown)


brianhelba:
@msmolens PTAL.

server/upload.py
```diff
@@ -180,7 +180,7 @@ def handleCsv(dataset, prereviewFolder, user, csvFile):
                     filename)
                 continue
             else:
-                imageItem = imageItems.next()
+                imageItem = next(imageItems)
```
msmolens: This change gives me an error:

      File "/home/vagrant/girder/plugins/isic_archive/server/upload.py", line 183, in handleCsv
        imageItem = next(imageItems)
    TypeError: MongoProxy object is not an iterator

Perhaps Girder's `MongoProxy` doesn't properly implement the iterator methods? https://docs.python.org/2/library/stdtypes.html#iterator-types

PR #308, opened 2016-12-02
Add a privateMetadata field to images, only visible to dataset reviewers

brianhelba:


brianhelba:
@msmolens PTAL.

server/__init__.py
```diff
@@ -87,16 +86,6 @@ def validateSettings(event):
...
 
 
-def onGetItem(event):
-    User = ModelImporter.model('user', 'isic_archive')
-    itemResponse = event.info['returnVal']
-
-    # Hide the 'originalFilename' metadata on Images from non-site admins
-    if 'originalFilename' in itemResponse.get('meta', {}):
-        if not User.canReviewDataset(getCurrentUser()):
-            del itemResponse['meta']['originalFilename']
```
msmolens: Is there a database migration in conjunction with this PR that removes `meta.originalFilename` from existing images?

brianhelba: Yes, I'll migrate this on the live site. Local test environments will have to be done separately, or just have a new batch of test images uploaded.

PR #309, opened 2016-12-04
Fix an issue where Mongo cursors are iterables but not iterators

brianhelba:


brianhelba:
@msmolens FYI. I'm merging this now to get it fixed.

PR #310, opened 2016-12-05
Add dataset metadata validation

msmolens:
- Move metadata upload into its own view.
- On the metadata upload view, user selects a dataset, uploads metadata, then can validate metadata for that dataset and save the metadata if validation succeeds.
- Parse errors and validation errors are displayed in separate sections.
- Preliminary validation routines for age and sex are included.


web_external/js/views/body/UploadDatasetMetadataView.js
```diff
@@ -0,0 +1,339 @@
...
+        // Prepare upload folder in user's home and start upload
+        if (this.uploadFolder) {
+            // Folder already created
+            this.startUpload(this.uploadFolder);
+        } else {
+            // Create new upload folder with unique name
+            this.uploadFolder = new girder.models.FolderModel({
+                name: 'isic_upload_' + Date.now(),
+                parentType: 'user',
+                parentId: girder.currentUser.get('_id'),
```
brianhelba: Should be `.id`.

brianhelba:
@msmolens I've just rebased this on the latest master, to account for all the moved file paths.

There should be no need to rebase it again, so feel free to put changes on top of this now.

web_external/js/views/body/UploadDatasetMetadataView.js
```diff
@@ -0,0 +1,339 @@
...
+        } else if (_.isEmpty(this.uploadedCsvFiles)) {
+            errorMessage = 'Please upload a CSV file.';
+        }
+
+        if (errorMessage !== null) {
+            isic.showAlertDialog({ text: errorMessage });
+            return;
+        }
+
+        this.dataset.off().on('isic:validated', function (resp) {
```
msmolens: I wonder if this pattern should be retired in favor of `listenTo()`? Even if `once()` is used here, there are two event handlers: one for success and one for error. At most one of those will occur. Does the remaining handler cause the model to leak?

msmolens:
I'd still like to replace where the .csv file is read directly from the assetstore, as well as rework the ProgressContext.

server/upload.py
```diff
@@ -144,29 +144,39 @@ def __init__(self, dataset, prereviewFolder, csvFile, user, validator):
...
+    def lines(self, stream):
+        """Generate individual lines of text from a stream."""
+        lastLine = ''
+        try:
+            # Read chunk from stream and split into lines. Always process the
+            # last line with the next chunk, or at the end of the stream,
+            # because it may be incomplete.
+            while True:
+                chunk = lastLine + ''.join(next(stream))
+                lines = chunk.splitlines(True)
```
brianhelba: Could we just use the keyword argument `keepends=True`?

web_external/Datasets/UploadDatasetMetadataView.js
```diff
@@ -319,7 +319,7 @@ isic.views.UploadDatasetMetadataView = isic.View.extend({
             return;
         }
 
-        this.dataset.off().on('isic:validated', function (resp) {
+        this.dataset.off().once('isic:validated', function (resp) {
```
brianhelba: Maybe refactor this to do the binding once in `initialize` using `listenTo`?

server/upload.py
msmolens: In Python 2 it looks like string functions don't take keyword arguments. It's in Python 3, but I'm not sure in which version it was added.

server/upload.py
```diff
@@ -124,32 +121,53 @@ def _fallbackUnzip(self):
...
+        # Metadata stored after validation, indexed by image item id
+        self.validatedMetadata = {}
 
-    assetstore = Assetstore.getCurrent()
-    assetstoreAdapter = assetstore_utilities.getAssetstoreAdapter(assetstore)
-    fullPath = assetstoreAdapter.fullPath(csvFile)
+        # Validation result, created after calling validate()
+        self.validationResult = None
+
+    def lines(self, stream):
```
brianhelba: This code should be extracted and made into it's own PR.

brianhelba:
Obsoleted by other PRs.

PR #312, opened 2016-12-08
Refactor segmentations to support reviews and mask format

brianhelba:


brianhelba:
@msmolens PTAL.

This updates the server to support all annotations being stored in mask form. A few notes about the transition:
* A server-side migration is now in progress to convert all existing segmentation to masks
* There are now API endpoints for setting and returning "review" information, but nothing in the client for this yet. It's more important to get the API working now, but we should add the client capability soon.
* The existing client still submits new segmentations as JSON, but they're immediately converted to masks as soon as they're received. We'll move the mask-creation logic to the client when it's updated to use GeoJS.

PR #313, opened 2016-12-08
Fix a bug in how the local feature array is created

brianhelba:
This does not affect the accuracy of any existing annotations, since JS will always include the final array element if it's actually set. Existing annotations should be migrated to add an extra "0" value, to make the new validation pass.


brianhelba:
Merging this as soon as tests pass.

PR #314, opened 2016-12-09
Add a "name" query string filter to the "/image" API endpoint

brianhelba:
Also harden security of model "find" methods.


brianhelba:
@aadikalloo This should help when translating an image name to it's ObjectId.

PR #316, opened 2016-12-09
Only expose the "Approximate Age" as clinical metadata

brianhelba:
The real age will be hidden as private metadata.


brianhelba:
Ping @aadikalloo : this means the field in the API will also change its name.

PR #318, opened 2016-12-13
Add text and links about 2017 challenge

brianhelba:


brianhelba:
Putting the styles directly into the template is a little bit of a hack, but this front page banner is going to be up temporarily.

PR #321, opened 2017-02-02
Miscellaneous JS fixes and cleanup

brianhelba:
See individual commit messages for details.

### DIFF
--- a/web_external/ImagesGallery/index.js
+++ b/web_external/ImagesGallery/index.js
@@ -68,6 +68,6 @@ isic.views.ImagesView = isic.View.extend({
     }
 });
 
-isic.router.route('images', 'images', function (id) {
+isic.router.route('images', 'images', function () {
     girder.events.trigger('g:navigateTo', isic.views.ImagesView);
 });
--- a/web_external/collection.js
+++ b/web_external/collection.js
@@ -1,0 +1,72 @@
+isic.Collection = girder.Collection.extend({ // eslint-disable-line backbone/collection-model
+
+    fetch: function (params, reset) {
+        // If "reset" is false, then reuse any unspecified params from the last
+        // fetch; "reset" defaults to true (so undefined will still reset)
+        if (reset === false) {
+            this.params = _.extend(this.params || {}, params);
+        } else {
+            this.params = params || {};
+        }
+
+        // Upstream Girder contains a bug where parameters are not honored on a
+        // reset fetch, so set the parameters manually before triggering a fetch
+        // with (ignored) params set to null, and the reset flag set to true.
+        // Doing a non-reset fetch causes the collection to not be emptied on
+        // an empty fetch, and does non-intuitive things with the "this.offset"
+        // property.
+        girder.Collection.prototype.fetch.call(this, null, true);
+    },
+
+    _currentOffset: function (params) {
+        // "this.offset" is not reliable, especially since we're doing a reset
+        // fetch in the upstream call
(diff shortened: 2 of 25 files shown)


web_external/collection.js
```diff
@@ -0,0 +1,71 @@
...
+            offset -= this.pageLimit;
+            offset = Math.max(0, offset);
+        }
+        this.fetch(_.extend(params, {offset: offset}), false);
+    },
+
+    pageNum: function () {
+        return Math.floor(this._currentOffset() / this.pageLimit);
+    }
+});
```
msmolens: My only concern is managing this if the plugin is upgraded for Girder 2.0.

web_external/ImagesGallery/Filter/ImagesFacetView.js
```diff
@@ -246,7 +246,7 @@ isic.views.ImagesFacetHistogramView = isic.views.ImagesFacetView.extend({
         knob.call(d3.behavior.drag()
             .origin(_.bind(function () {
                 return { x: 0, y: knobScale(this.scale.yMax) };
-            }, this)).on('drag', _.bind(function () {
+            }, this)).on('drag', function () {
```
msmolens: I don't think these d3 callbacks support passing a context.

web_external/ImagesGallery/index.js
```diff
@@ -24,7 +24,7 @@ isic.views.ImagesView = isic.View.extend({
             parentView: this
         });
 
-        $(window).on('resize.ImagesView', _.bind(this.render, this));
+        $(window).on('resize.ImagesView', this.render, this);
```
msmolens: jQuery event handlers don't support passing a context like this.

web_external/common/Viewer/ImageFullscreenWidget.js
```diff
@@ -2,17 +2,17 @@ isic.views.ImageFullscreenWidget = isic.View.extend({
...
             model: this.model
-        })).girderModal(this).on('shown.bs.modal', _.bind(function () {
+        })).girderModal(this).on('shown.bs.modal', function () {
             this.imageViewerWidget = new isic.views.ImageViewerWidget({
                 el: this.$('.isic-image-fullscreen-container'),
                 model: this.model,
                 parentView: this
             }).render();
-        }, this)).on('hidden.bs.modal', _.bind(function () {
+        }, this).on('hidden.bs.modal', function () {
```
msmolens: jQuery event handlers don't support passing a context like this.

web_external/ImagesGallery/Filter/ImagesFacetView.js
brianhelba: You're right, this was a find-replace oversight.

web_external/ImagesGallery/index.js
brianhelba: Fixing.

web_external/common/Viewer/ImageFullscreenWidget.js
brianhelba: Fixing.

web_external/collection.js
brianhelba: True, we may have to rewrite some of this for a migration, but having to individually hack every call to `fetch` is becoming unmaintainable, particularly when we want to use pagination with other potentially-changing parameters.

Eventually, I'd like to upstream some of these fixes, but that will have to be after the 2.0 migration.

PR #324, opened 2017-02-09
Send email confirmation when dataset is uploaded

### LINKED ISSUE #323: Send a confirmation email to the uploader of new datasets
Whenever a new dataset is uploaded (a request is made to the `POST /dataset` endpoint), an email should be sent to the user that just performed the upload. This email should contain all of the relevant dataset-level metadata that was just submitted (dataset name, description, license, attribution, etc.), as well as a hyperlink back to the Terms of Use page, which the user had agreed to on the upload form.

Possibly, we may want to send a copy of this email to the members of the "Dataset QC Reviewers" group and the admins of the "Dataset Contributers" group. Alternatively, we could send an abridged version of this email to those groups, without the full set of extra metadata.

This functionality could live either in [the API](https://github.com/ImageMarkup/isic-archive/blob/dd260802b4ae83f4a164fa69fb123911673f2c7e/server/api/dataset.py#L108) or [the model](https://github.com/ImageMarkup/isic-archive/blob/dd260802b4ae83f4a164fa69fb123911673f2c7e/server/models/dataset.py#L147) layers. We should look at where our other emails are triggered from, and consider that we may occasionally may want to ingest datasets directly via the model layer (in a server-side utility script), as well as consider the implications for code abstraction and testability.

msmolens:
Send an email confirmation once a dataset is uploaded. The email
includes all the dataset information that the user supplied on the
upload page, as well as a link to the agreed upon Terms of Use document.

Additionally, send an email to the 'Dataset QC Reviewers' group to
notify them about the new dataset. This email also includes the dataset
information.

Fixes #323


server/mail_templates/ingestDatasetConfirmation.mako
```diff
@@ -0,0 +1,18 @@
+<%include file="_header.mako"/>
+
+% if group:
+<%include file="ingestDatasetConfirmationHeaderGroup.mako"/>
```
brianhelba: Does it make sense to template these headers separately? I don't think they'll ever be reusable components.

server/mail_templates/ingestDatasetConfirmation.mako
```diff
@@ -0,0 +1,18 @@
...
+% else:
+<%include file="ingestDatasetConfirmationHeaderUser.mako"/>
+% endif
+
+<%include file="ingestDatasetConfirmationTable.mako"/>
```
brianhelba: Same here. Why not inline the table?

brianhelba:
LGTM.

PR #326, opened 2017-02-10
Obfuscate user names

brianhelba:


brianhelba:
@msmolens Ready for review.

brianhelba:
The user IDs are intended to be provided to everyone, as they're essential to do any sort of filtering or totally reliable analysis. For anonymous users, they should get a user ID and obfuscated user name in place of any user.

The group member listing issue is a bug in upstream Girder, due to the fact that it still (even in 2.0) does not fully support private users. For now, I think we'll be fine, given that the Girder API is undocumented and hidden. I'll get around to fixing this upstream some day, as I need support for private users for other projects too.

@msmolens Does that address your comments?

msmolens:
@brianhelba I meant that anonymous users can piece together real names from user ids by looking in the Annotation Studies collection.

brianhelba:
@msmolens Ah, I see what you mean. I'll fix in https://github.com/ImageMarkup/isic-archive/issues/328 , and do a database migration.

PR #329, opened 2017-02-14
Add Dataset listing view

msmolens:
- Add Dataset listing view
- Add pagination to listing views


brianhelba:
Also, once https://github.com/ImageMarkup/isic-archive/issues/331 is fixed, we should update this UI to only display those fields if they're returned by the API. However, the current public datasets do not have those fields set, so it's less critical.

msmolens:
The default limit in the client-side collections is [25](https://github.com/girder/girder/blob/599a9dbe7180f3f02bb7605205ec1c88f92fd368/clients/web/src/collection.js#L14), so that's plausible. We might even consider lowering that to make the listing views more usable.

brianhelba:
Ah, you're right.

PR #334, opened 2017-02-15
Remove metadata upload from upload dataset view

msmolens:


server/models/dataset.py
```diff
@@ -351,3 +326,17 @@ def reviewImages(self, dataset, acceptedImages, flaggedImages, user):
         if (Folder.countItems(prereviewFolder) +
                 Folder.countFolders(prereviewFolder)) == 0:
             Folder.remove(prereviewFolder)
+
+    def registerMetadata(self, dataset, csvFile):
+        """Register a .csv file containing metadata about images."""
+        # Add image metadata information to list
+        registeredMetadata = dataset['meta'].get('metadata', [])
```
brianhelba: Let's call this `metadataFiles`.

plugin_tests/upload_test.py
```diff
@@ -88,6 +88,27 @@ def testUpload(self):
...
+                'description': 'A test dataset',
+                'license': 'CC-0',
+                'signature': 'Test Uploader',
+                'anonymous': False,
+                'attribution': 'Test Organization'
+            })
+        self.assertStatusOk(resp)
+        dataset = resp.json
+
+        # Clean folder contents
```
brianhelba: Put this into another folder, instead of clearing and re-uploading to the first one.

server/api/dataset.py
```diff
@@ -242,3 +248,55 @@ def submitReviewImages(self, dataset, params):
 
         # TODO: return value?
         return {'status': 'success'}
+
+    @describeRoute(
+        Description('Register metadata with a dataset.')
+        .param('id', 'The ID of the dataset.', paramType='path')
+        .param('uploadFolderId', 'The ID of the folder that contains metadata '
```
brianhelba: Change this into `metadataFileId`, and have it point to a single Girder file, which should greatly simplify the validation here.

server/api/dataset.py
```diff
@@ -89,8 +96,7 @@ def getDataset(self, dataset, params):
 
     @describeRoute(
         Description('Create a lesion image dataset.')
-        .param('uploadFolderId', 'The ID of the folder that contains images '
-               'and metadata.')
+        .param('uploadFolderId', 'The ID of the folder that contains images.')
```
brianhelba: Change this into `uploadFileId`, and have it point to a single Girder file, which should greatly simplify the validation here (and in the `Dataset.ingestDataset` method).

server/api/dataset.py
```diff
@@ -242,3 +248,55 @@ def submitReviewImages(self, dataset, params):
...
+        csvFiles = Item.childFiles(csvFileItems[0], limit=2)
+        if not csvFiles.count():
+            raise ValidationException(
+                'No .csv files were uploaded.', 'uploadFolder')
+        if csvFiles.count() > 1:
+            raise ValidationException(
+                'Only one .csv file may be uploaded.', 'uploadFolder')
+        csvFile = csvFiles[0]
+
+        return Dataset.registerMetadata(dataset, csvFile)
```
brianhelba: This should email members of the `Dataset QC Reviewers` group, notifying them that new metadata is available.

msmolens:
@brianhelba PTAL. There is now a view to select a dataset, upload a metadata file, and register the metadata file with the dataset.

server/api/dataset.py
```diff
@@ -113,26 +120,30 @@ def getDataset(self, dataset, params):
...
-            uploadFolderId, user=user, level=AccessType.WRITE, exc=False)
-        if not uploadFolder:
+                'No file was uploaded.', 'zipFileId')
+        zipFile = File.load(
+            zipFileId, user=user, level=AccessType.WRITE, exc=False)
+        if not zipFile:
             raise ValidationException(
-                'Invalid upload folder ID.', 'uploadFolderId')
+                'Invalid upload file ID.', 'zipFileId')
+        if not mimetypes.guess_type(zipFile['name'], strict=False)[0] \
```
brianhelba: Let's make the algorithm this:
```
if (zipFile['mimeType'] not in ZIP_FORMATS) and \
   (zipFile['mimeType'] in ['application/octet-stream', None]) and\
   (mimetypes.guess_type(zipFile['name'], strict=False)[0] not in ZIP_FORMATS):
    raise ValidationException(...)
```

This accounts for the fact that files store their own MIME type in the `file['mimeType']` property. However, since that `mimeType` is set by the uploading client, we should fall back to type guessing if `mimeType` is a default or unknown value.

server/api/dataset.py
```diff
@@ -252,51 +263,36 @@ def submitReviewImages(self, dataset, params):
     @describeRoute(
         Description('Register metadata with a dataset.')
         .param('id', 'The ID of the dataset.', paramType='path')
-        .param('uploadFolderId', 'The ID of the folder that contains metadata '
-               'in a .csv file.')
+        .param('metadataFileId', 'The ID of the .csv metadata file.')
     )
     @access.user
-    @loadmodel(model='dataset', plugin='isic_archive', level=AccessType.READ)
+    @loadmodel(model='dataset', plugin='isic_archive', force=True)
```
brianhelba: Why is this? I would think that any user uploading metadata should at least be able to `READ` the dataset.

server/api/dataset.py
```diff
@@ -252,51 +263,36 @@ def submitReviewImages(self, dataset, params):
...
-            raise ValidationException(
-                'No .csv files were uploaded.', 'uploadFolder')
-        if csvFiles.count() > 1:
-            raise ValidationException(
-                'Only one .csv file may be uploaded.', 'uploadFolder')
-        csvFile = csvFiles[0]
-
-        return Dataset.registerMetadata(dataset, csvFile)
+                'Invalid metadata file ID.', 'metadataFileId')
+        if not mimetypes.guess_type(metadataFile['name'], strict=False)[0] \
```
brianhelba: I'd do the same checking here as with my suggestion for `ZIP_FORMATS`.

server/api/dataset.py
brianhelba: Combined with the `User.requireCreateDataset(user)` requirement below, I think this should be enough access checking. I don't think we can expect that all metadata uploaders will have `WRITE` access, since we may want to quickly take away end users' ability to delete items or arbitrarily modify metadata directly.

server/models/dataset.py
```diff
@@ -327,16 +296,49 @@ def reviewImages(self, dataset, acceptedImages, flaggedImages, user):
...
+                    'dataset': dataset,
+                    'user': user,
+                    'csvFile': csvFile,
+                    'date': now.replace(microsecond=0)
+                },
+                subject='ISIC Archive: Dataset Metadata Notification')
+
         return dataset
+
+    def _mailGroup(self, groupName, templateFilename, templateParams, subject):
```
brianhelba: This is a nifty helper. It might be nice to move it somewhere more general. I don't know where yet.

server/models/dataset.py
```diff
@@ -327,16 +296,49 @@ def reviewImages(self, dataset, acceptedImages, flaggedImages, user):
...
+        """Send email to all members of a group."""
+        Group = self.model('group')
+
+        group = Group.findOne({'name': groupName})
+        if not group:
+            raise GirderException('Could not load group: %s.' % groupName)
+        emails = [member['email'] for member in Group.listMembers(group)]
+        if emails:
+            html = mail_utils.renderTemplate(templateFilename, templateParams)
+            mail_utils.sendEmail(to=emails, subject=subject, text=html)
```
brianhelba: I'm debating whether we should loop over members here and send single emails directly to users, to avoid compromising the privacy of our users' email addresses.

Since the groups we're sending emails to now are internal administrators, for now, I think we should actually keep it as one email with multiple recipients, as it could be convenient for a recipient to Reply-All and notify other group members of something. If we start sending mails to "external" user group, like the entirety of `Dataset Contributors`, then we should consider the looping approach, or maybe add a function parameter to control single vs mass emails.

server/models/dataset.py
```diff
@@ -327,16 +296,49 @@ def reviewImages(self, dataset, acceptedImages, flaggedImages, user):
...
+        if self.findOne({'meta.metadataFiles.fileId': csvFile['_id']}):
+            raise ValidationException('Metadata file is already registered.')
+
+        # Add image metadata file information to list
+        now = datetime.datetime.utcnow()
+        metadataFiles = dataset['meta'].get('metadataFiles', [])
+        metadataFiles.append({
             'fileId': csvFile['_id'],
-            'time': datetime.datetime.utcnow()
+            'user': user['_id'],
```
brianhelba: Yeah, `user` is an important addition. Let's make it `userId`, just for consistency with stuff like `creatorId`.

server/models/dataset.py
```diff
@@ -327,16 +296,49 @@ def reviewImages(self, dataset, acceptedImages, flaggedImages, user):
...
 
-    def registerMetadata(self, dataset, csvFile):
+    def registerMetadata(self, dataset, csvFile, user, sendMail=False):
         """Register a .csv file containing metadata about images."""
-        # Add image metadata information to list
-        registeredMetadata = dataset['meta'].get('metadata', [])
-        registeredMetadata.append({
+        # Check if image metadata is already registered
+        if self.findOne({'meta.metadataFiles.fileId': csvFile['_id']}):
+            raise ValidationException('Metadata file is already registered.')
```
brianhelba: Since we're checking all datasets globally, let's make the error message `'Metadata file is already registered on a dataset.'`

web_external/models/ItemModel.js
```diff
@@ -0,0 +1,15 @@
+isic.models.ItemModel = girder.models.ItemModel.extend({
+    /**
+     * Get the files within the item.
+     */
+    getFiles: function () {
```
brianhelba: Since `Item` is not typically something we deal with directly in ISIC, I think it'd be better to patch this method onto the `girder.models.ItemModel` instead (or maybe the prototype, I forget how Backbone inheritance works). Eventually, we should not need this code anyway, once the uploader widget does what we want it to.

web_external/models/FolderModel.js
```diff
@@ -0,0 +1,15 @@
+isic.models.FolderModel = girder.models.FolderModel.extend({
+    /**
+     * Remove the contents of the folder.
+     */
+     removeContents: function () {
```
brianhelba: Again, I'd rather not confuse our ISIC models with having a Girder thing here. I'd either patch this onto the `girder.models.FolderModel`, or since this code may be more permanent than the `ItemModel.getFiles` make it into a utility method on the View, whichever you think is better.

web_external/collections/ItemCollection.js
```diff
@@ -0,0 +1,3 @@
+isic.collections.ItemCollection = girder.collections.ItemCollection.extend({
```
brianhelba: Per my comments about `isic.models.ItemModel`, I'd prefer this go away.

web_external/models/ItemModel.js
brianhelba: Of course, if you feel differently, let's discuss it.

web_external/Datasets/UploadDatasetView.js
```diff
@@ -125,49 +166,18 @@ isic.views.UploadDatasetView = isic.View.extend({
         this.uploadWidget.uploadNextFile();
     },
 
-    submitDataset: function () {
+    submitDataset: function (zipFileId) {
```
brianhelba: To be clear, if `zipFileId === null`, we make a REST request anyway, and let the server be the one to determine it's an error? I don't actually see anything wrong with this, but maybe make that clear in a code comment.

brianhelba:
@msmolens Ugh, GitHub just ate a bunch of my comments by marking them as outdated. You should still be able to see them on this page, so feel free to respond inline there, or we can discuss in-person.

web_external/layout/layoutHeader.jade
```diff
@@ -82,6 +82,11 @@ nav.navbar.navbar-default.navbar-fixed-top.isic-header-wrapper
                 a(title="Upload Dataset", href="#uploadDataset")
                   i.icon-upload
                   | Upload Dataset
+            if currentUser
```
brianhelba: Let's only render this if the current user has upload permissions too. There no need to confuse most users with 2 places to get to the form to request contributor access.

web_external/Datasets/UploadDatasetView.js
```diff
@@ -42,19 +30,59 @@ isic.views.UploadDatasetView = isic.View.extend({
...
         });
 
+        this.dataset = new isic.models.DatasetModel();
+
+        this.listenTo(this.dataset, 'isic:ingestImages:success', function () {
+            isic.showAlertDialog({
+                text: '<h4>Dataset successfully submitted.</h4>',
+                escapedHtml: true,
+                callback: function () {
+                    isic.router.navigate('', {trigger: true});
```
brianhelba: At this point, we should navigate straight to the upload dataset metadata view.

web_external/Datasets/UploadDatasetMetadataView.js
```diff
@@ -0,0 +1,248 @@
+// View for a collection of datasets in a select tag
+isic.views.UploadDatasetMetadataSelectDatasetView = isic.View.extend({
```
brianhelba: Sorry to do this, but I actually think we should ditch the entire dataset selector box in favor of making this view specific to a single dataset. We could update the routing to accept a dataset ID parameter (which would then be passed to the `UploadDatasetMetadataView`.

I think this has several advantages:
* The `UploadDatasetView` can redirect people directly to a view which adds metadata only to the dataset they just uploaded, which reduces confusion and decision points for newer contributers.
* Since we're only requiring `READ` permissions to upload to a dataset, we should attempt to prevent users from accidentally contributing metadata directly to a public datasetl
* In the future, we may direct users to this dialog via buttons dataset listing view, so we'll be able to go directly to a known dataset ID.

web_external/layout/layoutHeader.jade
brianhelba: On second thought, I think this menu item should go away entirely. Users can get to this view via either:
* a redirect after uploading
* the dataset listing page

web_external/Datasets/UploadDatasetMetadataView.js
brianhelba: Again, we can discuss if you feel differently.

web_external/Datasets/UploadDatasetMetadataView.js
```diff
@@ -0,0 +1,248 @@
...
+        select.select2({
+            placeholder: placeholder
+        });
+        select.focus();
+
+        return this;
+    }
+});
+
+isic.views.UploadDatasetMetadataView = isic.View.extend({
```
brianhelba: This `UploadDatasetMetadataView` is the only thing in this PR that I haven't reviewed yet, just because it could change substantially if we get rid of the `UploadDatasetMetadataSelectDatasetView`.

brianhelba:
@msmolens Ok, finished my review. This was a really impressive amount of additional functionality for you to add in a single day! Let's discuss in more detail ASAP.

As I mentioned, forgive the fact that some of my comments are marked as outdated. You'll have to expand them from this page.

msmolens:
@brianhelba PTAL, I believe I addressed all the feedback.

web_external/Datasets/UploadDatasetMetadataView.js
```diff
@@ -0,0 +1,195 @@
...
```
brianhelba: Let's rename this View, and all associated JS, CSS, and Jade files (and probably some HTML classes and IDs :slightly_frowning_face: ) to be something like `RegisterMetadataDatasetView`, as:
* we're referring to this process internally on the server-side as "registering" metadata, rather than just "uploading" it
* the ordering of words makes the place of this functionality within the logical site hierarchy more clear

web_external/Datasets/UploadDatasetMetadataView.js
```diff
@@ -0,0 +1,195 @@
...
+        //  (2) Request and receive create dataset access
+        // before being able to see the upload dataset metadata view
+        if (!isic.models.UserModel.currentUserCanAcceptTerms()) {
+            girder.events.trigger('g:navigateTo', isic.views.TermsAcceptanceView);
+        } else if (!girder.currentUser.canCreateDataset()) {
+            girder.events.trigger('g:navigateTo', isic.views.UploadDatasetRequestView);
+        } else {
+            // Fetch the dataset, then navigate to the view
+            var dataset = new isic.models.DatasetModel();
+            dataset.set({
```
brianhelba: Just curious, is there a reason not to pass the `_id` property directly to the `DatasetModel` constructor, on the previous line?

web_external/Datasets/UploadDatasetMetadataView.js
```diff
@@ -0,0 +1,195 @@
...
+    },
+    uploadStarted: function (files) {
+        // Prepare upload folder in user's home and start upload
+        if (this.uploadFolder) {
+            // Folder already created
+            this.startUpload(this.uploadFolder);
+        } else {
+            // Create new upload folder with unique name
+            this.uploadFolder = new girder.models.FolderModel({
+                name: 'isic_upload_' + Date.now(),
```
brianhelba: Let's prefix this folder name as `isic_metadata_` instead, just to make it easier if we move these around later.

web_external/Datasets/UploadDatasetMetadataView.js
```diff
@@ -0,0 +1,195 @@
...
+                parentType: 'user',
+                parentId: girder.currentUser.id,
+                description: 'ISIC uploads'
```
brianhelba: And maybe change the description, but it doesn't actually matter.

web_external/Datasets/UploadDatasetMetadataView.js
```diff
@@ -0,0 +1,195 @@
...
+            this.uploadFolder.once('g:saved', function () {
+                this.startUpload(this.uploadFolder);
+            }, this).once('g:error', function () {
+                girder.events.trigger('g:alert', {
+                    icon: 'cancel',
+                    text: 'Could not create upload folder.',
+                    type: 'error',
+                    timeout: 4000
+                });
+            }, this).save();
```
brianhelba: Why is the folder getting `.save()` called on it here?

brianhelba:
Looks great, and I've just successfully tested it locally.

I'll just await your responses to my last minor bits of feedback.

web_external/Datasets/UploadDatasetMetadataView.js
msmolens: Would `RegisterMetadataView` be descriptive enough? I have trouble parsing the "Dataset" part of `RegisterMetadataDatasetView` when it's not describing the metadata (i.e. "dataset metadata").

web_external/Datasets/UploadDatasetMetadataView.js
msmolens: Calling `.save()` runs the request that creates the folder on the server:
https://github.com/girder/girder/blob/1.x-maintenance/clients/web/src/model.js#L32-L37

web_external/Datasets/UploadDatasetMetadataView.js
brianhelba: Yeah, I think that's better, actually.

web_external/Datasets/UploadDatasetMetadataView.js
brianhelba: Oh, I just misread this block. Nevermind.

web_external/Datasets/UploadDatasetMetadataView.js
msmolens: I agree it's nicer to do that; I updated the code. The only issue is that effectively makes `dataset` an unused variable, since it's used only in the chained function calls, but this passes the style check. I believe omitting the variable assignment would necessitate manually disabling a style check for that line, which is uglier IMO.

web_external/Datasets/UploadDatasetMetadataView.js
msmolens: I updated this and the folder name in the upload dataset view for consistency.

web_external/Datasets/UploadDatasetMetadataView.js
msmolens: Makes sense.

web_external/Datasets/UploadDatasetMetadataView.js
brianhelba: `dataset` is passed to the new view as an argument.

msmolens:
@brianhelba Great. Thanks for all the reviews!

brianhelba:
This branch is now live too. :fireworks:

PR #342, opened 2017-02-22
Apply registered metadata on the server

brianhelba:
Before being merged, this will require a database update on the live server, for commit https://github.com/ImageMarkup/isic-archive/pull/342/commits/88f25acefd6634837b6aae9079cb689bc7708e23

### DIFF
--- a/plugin_tests/upload_test.py
+++ b/plugin_tests/upload_test.py
@@ -278,13 +278,13 @@ class UploadTestCase(IsicTestCase):
         self.assertStatus(resp, 401)
         resp = self.request(
             path='/dataset/%s/metadata' % publicDataset['_id'], method='GET',
-            user=reviewerUser)
+            user=uploaderUser)
         self.assertStatus(resp, 403)
 
         # List (empty) registered metadata
         resp = self.request(
             path='/dataset/%s/metadata' % publicDataset['_id'], method='GET',
-            user=uploaderUser)
+            user=reviewerUser)
         self.assertStatusOk(resp)
         self.assertEqual(resp.json, [])
 
@@ -302,7 +302,7 @@ class UploadTestCase(IsicTestCase):
         # List registered metadata
         resp = self.request(
             path='/dataset/%s/metadata' % publicDataset['_id'],
-            user=uploaderUser)
+            user=reviewerUser)
         self.assertStatusOk(resp)
         self.assertIsInstance(resp.json, list)
         self.assertEqual(len(resp.json), 1)
--- a/plugin_tests/upload_test.py
+++ b/plugin_tests/upload_test.py
@@ -319,6 +319,6 @@ class UploadTestCase(IsicTestCase):
                 '_id': str(uploaderUser['_id']),
                 'name': User.obfuscatedName(uploaderUser)
             },
-            # This is is actually checked above
+            # This is actually checked above
             'time': resp.json[0]['time']
         }, resp.json[0])
--- a/plugin_tests/upload_test.py
+++ b/plugin_tests/upload_test.py
@@ -54,6 +54,9 @@ class UploadTestCase(IsicTestCase):
             'mongodb://localhost:27017/girder_worker')
         # TODO: change this to 'amqp://guest@127.0.0.1/' for RabbitMQ
 
+        self.testDataDir = os.path.join(
+            os.environ['GIRDER_TEST_DATA_PREFIX'], 'plugins', 'isic_archive')
+
     def _uploadDataset(self, uploaderUser, zipName, zipContentNames,
                        datasetName, datasetDescription):
         Dataset = self.model('dataset', 'isic_archive')
@@ -61,12 +64,11 @@ class UploadTestCase(IsicTestCase):
         Upload = self.model('upload')
 
         # Create a ZIP file of images
-        testDataDir = os.path.join(
-            os.environ['GIRDER_TEST_DATA_PREFIX'], 'plugins', 'isic_archive')
         zipStream = BytesIO()
         zipGen = ZipGenerator(zipName)
         for fileName in zipContentNames:
-            with open(os.path.join(testDataDir, fileName), 'rb') as fileObj:
+            with open(os.path.join(self.testDataDir, fileName), 'rb') as \
+                    fileObj:
                 for data in zipGen.addFile(lambda: fileObj, fileName):
                     zipStream.write(data)
         zipStream.write(zipGen.footer())
@@ -120,7 +122,10 @@ class UploadTestCase(IsicTestCase):
         return dataset
 
     def testUploadDataset(self):
+        File = self.model('file')
+        Folder = self.model('folder')
         Group = self.model('group')
+        Upload = self.model('upload')
         User = self.model('user', 'isic_archive')
 
         # Create a reviewer user that will receive notification emails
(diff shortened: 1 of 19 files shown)


brianhelba:
@msmolens I've rebased this again, to incorporate new the `IsicResource._decodeParams` method.

server/api/dataset.py
```diff
@@ -248,6 +250,33 @@ def submitReviewImages(self, dataset, params):
...
+        User = self.model('user', 'isic_archive')
+
+        user = self.getCurrentUser()
+        User.requireCreateDataset(user)
+
+        return [
+            {
+                'file': File.load(
+                    registration['fileId'], force=True, exc=True,
+                    fields=['_id', 'name']),
```
msmolens: Specifying `fields` causes an error in the Girder version we use:
```
  File "/home/vagrant/girder/girder/utility/acl_mixin.py", line 63, in load
    loadType = doc['attachedToType']
KeyError: 'attachedToType'
```
To fix it I think we'd need https://github.com/girder/girder/commit/936b0ba9ea10264e5a6f0bae1516d804fd061cfd and perhaps the followup commit https://github.com/girder/girder/commit/8b10a0807214a14319682e9c6f9f809f7ccac538.

server/models/dataset.py
```diff
@@ -326,9 +328,124 @@ def registerMetadata(self, dataset, csvFile, user, sendMail=False):
...
+        Apply metadata in a .csv file to a dataset.
+        :return: A list of strings describing parsing or validation errors.
+        :rtype: list
+        """
+        Assetstore = self.model('assetstore')
+
+        assetstore = Assetstore.getCurrent()
+        assetstoreAdapter = assetstore_utilities.getAssetstoreAdapter(
+            assetstore)
+        metadataFilePath = assetstoreAdapter.fullPath(metadataFile)
```
msmolens: The abandoned PR (https://github.com/ImageMarkup/isic-archive/pull/310) includes a change to support general assetstores. We could pull a version of that in at some point.

brianhelba:
I think we should add a button to directly download the currently selected CSV file, as a convenience.

brianhelba:
We should also add a notation to the registered metadata, once it's been applied (and perhaps containing the application time),

msmolens:
I think the Apply metadata view should also show a Dataset info box like on the Register metadata view. I'll add this.

web_external/Datasets/ApplyMetadataView.js
```diff
@@ -0,0 +1,261 @@
+// Model for a metadata file
+isic.models.MetadataFileModel = girder.Model.extend({
+    name: function () {
+        var time = girder.formatDate(this.get('time'), girder.DATE_SECOND);
+        var fileName = this.get('file').name();
+        var userName = this.get('user').name();
+        return time + ' \u2014 ' + fileName + ' \u2014 ' + userName;
```
brianhelba: Why use a unicode "EM DASH", rather than a simple ASCII dash?

web_external/Datasets/ApplyMetadataView.js
```diff
@@ -0,0 +1,261 @@
...
+
+            // Use file ID as item ID
+            item._id = item.file.id;
+        });
+        return data;
+    }
+});
+
+// Model for a single metadata error
+isic.models.MetadataErrorModel = girder.Model.extend({
```
brianhelba: Any reason why this is a `girder` model, rather than a `Backbone` one?

web_external/Datasets/ApplyMetadataView.js
```diff
@@ -0,0 +1,261 @@
...
+    description: function () {
+        return this.get('description');
+    }
+});
+
+// Collection of metadata errors. The list of items in the collection is
+// meaningful only when initialized() is true.
+isic.collections.MetadataErrorCollection = girder.Collection.extend({
```
brianhelba: Any reason why this is a `girder` collection, rather than a `Backbone` one?

web_external/Datasets/ApplyMetadataView.js
msmolens: Em dash is much wider than a hyphen, and the role of the character here is to be a helpful separator between the components. A more advanced user interface here than a select box could be appropriate, and that would avoid using characters to separate information.

web_external/Datasets/ApplyMetadataView.js
msmolens: In this case, it's probably better as a `Backbone` model. I will update the code.

web_external/Datasets/ApplyMetadataView.js
msmolens: In this case, it's probably better as a `Backbone` collection. I will update the code.

web_external/Datasets/ApplyMetadataView.js
brianhelba: Sounds good.

brianhelba:
@msmolens Ok, I've made the changes that we discussed, fixed testing, and cleaned up a few other small areas.

I still need to finish reviewing the content of `ApplyMetadataView.js`, which I'll do now. Please take a look at my changes.

The actual server-side logic for applying metadata is still totally broken, but I'm fine with merging it as-is (maybe we'll just change the access decorator to be admin-only). We'll clean up the server logic in a subsequent PR soon.

plugin_tests/upload_test.py
```diff
@@ -204,3 +211,114 @@ def testUploadDataset(self):
...
+        self.assertDictEqual({
+            'file': {
+                '_id': str(metadataFile['_id']),
+                'name': metadataFile['name']
+            },
+            'user': {
+                '_id': str(uploaderUser['_id']),
+                'name': User.obfuscatedName(uploaderUser)
+            },
+            # This is is actually checked above
```
msmolens: Typo: duplicate "is"

web_external/Datasets/ApplyMetadataView.js
```diff
@@ -0,0 +1,261 @@
...
+
+        this.dataset.getRegisteredMetadata().done(_.bind(function (resp) {
+            this.files.reset(resp, {parse: true});
+        }, this));
+
+        this.render();
+    },
+
+    fileChanged: function (file) {
+        this.file.set(file.attributes);
```
brianhelba: In general, I'd like to move away from this sort of thing, where we make copies of models, rather than more globally sharing a single instance. Of course, I see why you did this here, and it may indeed be the simplest way to handle it.

I wonder about the possibilities of either:
* Move the instantiation of `ApplyMetadataValidationView` into a separate method which gets called (effectively) within `filesChanged`. The `ApplyMetadataValidationView` would take a `file` as a parameter, then be destroyed and re-instantiated every time the `file` selection changes.
* Have `ApplyMetadataValidationView` expose a method to change its own file, re-rendering itself when that occurs.

I'd like to talk this through a bit further, just to be certain we're doing things the best way here.

brianhelba:
@msmolens I've pushed a couple minor things, as well as a larger change in https://github.com/ImageMarkup/isic-archive/pull/342/commits/54b61d5e220b4f877134829fe83a25e4d53de601 . PTAL at that change, and if you have any objections, we can discuss and modify or revert it.

I have one additional minor question inline, but otherwise, this LGTM.

web_external/Datasets/applyMetadataPage.jade
```diff
@@ -0,0 +1,25 @@
...
+        |  Download
+    .isic-form-group
+      h3 Validate metadata
+      p.
+        This section displays any problems with the metadata.
+        Press the button below to validate the metadata.
+        Once validation succeeds, you will be given an option to save the metadata.
+      .isic-apply-metadata-validate-button-container.text-center
+        button#isic-apply-metadata-validate-button.btn.btn-primary.btn-md(disabled) Validate
+      .isic-apply-metadata-validation-container
```
brianhelba: Why is it necessary to have two levels of `<div>` here? Couldn't we just put the `#isic-apply-metadata-validation-container` at the upper level, and dispense with the `.isic-apply-metadata-validation-container` entirely?

web_external/Datasets/ApplyMetadataView.js
msmolens: I agree that there should be no need to duplicate models. I like your second solution above--adding a set method to triggers rerendering. However now that `ApplyMetadataValidationView` is a simple wrapper around a template, I thought it made sense to remove the view entirely and render that template in the top view. See https://github.com/ImageMarkup/isic-archive/pull/342/commits/bf6dd60acd83061df4354f6ea5fd24725043e846.

web_external/Datasets/ApplyMetadataView.js
```diff
@@ -143,6 +147,9 @@ isic.views.ApplyMetadataView = isic.View.extend({
         }
     },
 
+    /**
+     * @param {isic.models.DatasetModel} settings.dataset
+     */
```
msmolens: If we're feeling ambitious we could consider using JSDoc format consistently throughout the client code.

msmolens:
@brianhelba I pushed a few minor changes, and don't have any more planned.
Some improvements that could go in later could be:

- After applying metadata, redirect back to dataset view, open to the correct dataset, instead of the front page. This would also apply to the upload dataset and register metadata views.
- Consider showing metadata file information on the dataset page.

PR #344, opened 2017-02-23
Add an image download button to details pane, and refactor

brianhelba:


web_external/ImagesGallery/Detail/imageDetailsPage.jade
```diff
@@ -1,14 +1,33 @@
...
+        title='Open this image in a new window'
+        href='/api/v1/image/' + image.id + '/download?contentDisposition=inline'
+        target='_blank'
+      )
         i.icon-link-ext
+      a#isic-image-details-download.isic-image-details-button.btn.btn-xs.btn-default(
+        data-toggle='tooltip'
+        data-placement='bottom'
+        title='Download this image'
+        href='/api/v1/image/' + image.id + '/download'
```
msmolens: FYI I see a warning in Chrome:
```
Resource interpreted as Document but transferred with MIME type image/jpeg: "http://isic-archive.upstream:8080/api/v1/image/589de960d831136be37e0d8e/download".
```

This can be avoided by adding a `download` attribute to the anchor (described at https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a).

Also, regardless of the presence of that attribute I noticed that Firefox behaves differently in that it always presents a file save dialog. Chrome downloads the file immediately.

brianhelba: That's a good point. I actually thought about adding the attribute, but I didn't like having to redundantly add the file name there too. However, the docs also state
> If the HTTP header Content-Disposition: gives a different filename than this attribute, the HTTP header takes priority over this attribute.
so I can probably just add the attribute with no value. I'll try it now.

PR #345, opened 2017-02-23
Improve upload widget drop zone style

msmolens:
This commit improves the style applied to the upload widget, especially
with regards to drag and drop.

The drop zone is taller and wider. This makes it easier to drag and drop
to upload a file.

Additionally, the width of "Browse or drop files" button no longer
depends on the width of the text below (which changes after uploading a
file).

### DIFF
--- a/web_external/Datasets/uploadWidget.styl
+++ b/web_external/Datasets/uploadWidget.styl
@@ -9,7 +9,7 @@
     font-color #eee
 
   .g-drop-zone
-    width 100%
+    width 40%
     text-align center
     height 60px
     line-height 60px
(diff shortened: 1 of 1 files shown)


web_external/Datasets/uploadWidget.styl
```diff
@@ -30,17 +30,22 @@
       background #6cc8a7
       box-shadow inset 0 3px 5px rgba(0,0,0,0.125);
 
+    &.g-drop-zone
```
brianhelba: What's this selector do? The parent is also `.g-drop-zone`, so does it compile to be `.g-drop-zone.g-drop-zone`?

msmolens: That's an error, thanks.

msmolens:
Screenshot on Windows:
image

PR #346, opened 2017-02-24
Create API endpoints to add additional images and multiple users to a study

brianhelba:


server/api/study.py
```diff
@@ -300,31 +293,96 @@ def createStudy(self, params):
...
+        duplicateAnnotations = Annotation.find({
+            'meta.studyId': study['_id'],
+            'meta.imageId': {'$in': [image['_id'] for image in images]}
+        })
+        if duplicateAnnotations.count():
+            # Just list the first duplicate
+            duplicateAnnotation = next(iter(duplicateAnnotations))
+            raise ValidationException(
+                'Image "%s" is already part of the study.' %
+                duplicateAnnotation['meta']['imageId'])
```
msmolens: I was testing and found I can call this endpoint repeated with the same user ID without triggering this warning. I'd have to actually annotate an image before I'd get this warning, it looks like?

brianhelba: This should be fixed now.

PR #349, opened 2017-02-27
Add an API endpoint to download multiple original images in a ZIP file

brianhelba:


brianhelba:
I think this requires some more consideration.
* If we remove the requirement of specifying a `datasetId` parameter, I'm concerned that users may inadvertently DOS our computational resources or, more likely, our bandwidth budget, as they'll be downloading multiple GBs with a single, anonymous request.
* If we add support for a `filter` parameter, we can integrate this into the Images Gallery UI, which is definitely a goal. However, currently, the `filter` parameter is not documented quite complicated (due to the PUG-JS grammar it uses), so it's presently impractical for direct API users to compose requests that use `filter`.
* We probably want to also add support for something like an `updatedSince` parameter, which filters only images that have been added or updated since a given date. However, we need to make stronger internal guarantees about precisely when the `modified` time on images is getting updated. Additionally, for feature parity, the `GET /image` endpoint should also support this parameter.

brianhelba:
We need to include a CC-0 license file in this ZIP.

brianhelba:
@msmolens PTAL.

server/api/image.py
```diff
@@ -94,6 +98,84 @@ def find(self, params):
...
+                datasetId = image['folderId']
+                if datasetId not in datasetCache:
+                    datasetCache[datasetId] = Dataset.load(
+                        datasetId, force=True, exc=True)
+                dataset = datasetCache[datasetId]
+                imageFile = Image.originalFile(image)
+                imageFileGenerator = File.download(imageFile, headers=False)
+                for data in zipGenerator.addFile(
+                        imageFileGenerator,
+                        path=os.path.join(dataset['name'], imageFile['name'])):
```
msmolens: If the dataset name contains Unicode characters, the folder extracted from the zip file is unexpectedly expanded to the UTF-8 encoding of the name. Unsurprisingly a similar topic has come up before: https://github.com/girder/girder/issues/985

It looks like the zip spec was updated to include a "language encoding" flag that might help with this. See https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT, keywords: "Bit 11", "Appendix D".

Clearly this isn't something that should be further addressed in this PR.

server/api/image.py
```diff
@@ -94,6 +98,84 @@ def find(self, params):
...
+                    {
+                        'work': dataset['name'],
+                        'author':
+                            dataset['meta']['attribution']
+                            if not dataset['meta']['anonymous']
+                            else 'Anonymous'
+                    })
+                for data in zipGenerator.addFile(
+                        lambda: [licenseText],
+                        path=os.path.join(dataset['name'], 'license.txt')):
```
msmolens: Would it be better to capitalize this, like: `LICENSE.txt`?

server/api/image.py
```diff
@@ -94,6 +98,84 @@ def find(self, params):
...
+                    yield data
+                for data in zipGenerator.addFile(
+                        lambda: [attributionText],
+                        path=os.path.join(dataset['name'], 'attribution.txt')):
```
msmolens: Would it be better to capitalize this, like: `ATTRIBUTION.txt`?

server/api/image.py
```diff
@@ -94,6 +98,84 @@ def find(self, params):
...
+                imageFile = Image.originalFile(image)
+                imageFileGenerator = File.download(imageFile, headers=False)
+                for data in zipGenerator.addFile(
+                        imageFileGenerator,
+                        path=os.path.join(dataset['name'], imageFile['name'])):
+                    yield data
+
+            for dataset in six.viewvalues(datasetCache):
+                licenseText = mail_utils.renderTemplate(
+                    'license_%s.mako' % dataset['meta']['license'])
```
msmolens: I hit an error here:
```
CompileException: Unicode decode operation of encoding 'ascii' failed in file '/home/vagrant/girder/plugins/isic_archive/server/license_templates/license_CC-BY.mako' at line: 0 char: 
```
This likely affects CC-BY and CC-BY-NC, since they are not ASCII:
```
attribution_CC-0.mako:     ASCII text, with CRLF line terminators
attribution_CC-BY.mako:    ASCII text, with CRLF line terminators
attribution_CC-BY-NC.mako: ASCII text, with CRLF line terminators
license_CC-0.mako:         ASCII text, with CRLF line terminators
license_CC-BY.mako:        UTF-8 Unicode text, with CRLF line terminators
license_CC-BY-NC.mako:     UTF-8 Unicode text, with CRLF line terminators
```

server/api/image.py
brianhelba: Sure. I was already considering it.

server/api/image.py
brianhelba: Sure. I was already considering it.

server/api/image.py
brianhelba: Can you file a bug, to revisit this once we're on a newer version of Girder?

server/api/image.py
```diff
@@ -94,6 +98,84 @@ def find(self, params):
...
+                    query.update({'folderId': ObjectId(params['datasetId'])})
+                except InvalidId:
+                    raise RestException(
+                        'Invalid "folderId" ObjectId: %s' % params['datasetId'])
+
+        user = self.getCurrentUser()
+        downloadFileName = 'ISIC-images'
+
+        def stream():
+            datasetCache = dict()
```
brianhelba: Fixing globally.

server/api/image.py
```diff
@@ -94,6 +98,84 @@ def find(self, params):
...
+        .errorResponse()
+    )
+    @access.public
+    def downloadMultiple(self, params):
+        Dataset = self.model('dataset', 'isic_archive')
+        File = self.model('file')
+        Image = self.model('image', 'isic_archive')
+
+        if 'filter' in params:
+            query = querylang.astToMongo(json.loads(params['filter']))
```
brianhelba: I'll fix this everywhere.

readme.rst
```diff
@@ -110,7 +110,7 @@ Usage
     # Mongodb's console output
     tail -F /var/log/mongodb/mongod.log
 
-.. |build-status| image:: https://travis-ci.org/ImageMarkup/isic-archive.svg?branch=master
+.. |build-status| image:: https://travis-ci.org/ImageMarkup/isic-archive/master.svg
```
brianhelba: Fixing. The issue before was that the badge wasn't displaying the status of "master", but just the most recent build.

server/api/image.py
brianhelba: Fixed. These were copied directly from https://creativecommons.org/2014/01/07/plaintext-versions-of-creative-commons-4-0-licenses/ .

server/api/image.py
```diff
@@ -94,6 +98,84 @@ def find(self, params):
...
+                        path=os.path.join(dataset['name'], 'attribution.txt')):
+                    yield data
+
+            yield zipGenerator.footer()
+
+        setResponseHeader('Content-Type', 'application/zip')
+        setResponseHeader(
+            'Content-Disposition',
+            'attachment; filename="%s.zip"' % downloadFileName)
+        return stream
```
brianhelba: IMO, this is just a corner case, not an error. As long as it doesn't crash, I'd say it's fine to make an empty ZIP, to keep the return type in all non-error cases the same (`application/zip`).

web_external/ImagesGallery/PagingPane.js
```diff
@@ -6,7 +6,11 @@ isic.views.ImagesViewSubViews.PagingPane = Backbone.View.extend({
         'click #isic-images-seekFirst': 'seekFirst',
         'click #isic-images-seekPrev': 'seekPrev',
         'click #isic-images-seekNext': 'seekNext',
-        'click #isic-images-seekLast': 'seekLast'
+        'click #isic-images-seekLast': 'seekLast',
+        'click #isic-images-download-zip': function () {
+            window.location.assign(girder.apiRoot + '/image/download?filter=' +
+                JSON.stringify(this.model.getFilterAstTree()));
+        }
```
brianhelba: Good catch. The bug here is really the failure to apply the rule from https://github.com/ImageMarkup/isic-archive/pull/352 to this endpoint. Fixed now.

brianhelba:
@msmolens PTAL again. I addressed all of your comments. Thanks for the good review.

server/api/image.py
msmolens: Now if I browse to Lesion Images, don't filter anything, and click "Download as ZIP" I see a JSON error. This happens when `getFilterAstTree()` returns `undefined`, which is added literally to the URL: `http://localhost:8080/api/v1/image/download?filter=undefined`.

brianhelba:
@msmolens Should be all fixed now.

I previously wasn't understanding why you were passing `undefined` to the filter argument, but I fixed the place in the client where that was happening. I also fixed some mis-indented new code that caused filtering to be broken.

Finally, per your point about empty ZIP files, I've disabled the download button on the client when no images are selected. The API will still return empty ZIPs if you request it, though.

msmolens:
@brianhelba Makes sense. I tested all the new commits locally and things look good.

PR #350, opened 2017-02-28
Use obfuscated user names more comprehensively

### LINKED ISSUE #328: Remove user names from annotation study sub-folders
Annotator user names should be considered private. The naming of study sub-folders allows anyone with view access to a study to see the real name of the annotators.

brianhelba:
Fixes #328

### DIFF
--- a/web_external/models/UserModel.js
+++ b/web_external/models/UserModel.js
@@ -1,14 +1,24 @@
 isic.models.UserModel = girder.models.UserModel.extend({
     name: function () {
-        var name = this.get('name');
+        var realName;
         if (this.has('login')) {
-            var privateName =
+            realName =
                 this.get('firstName') +
                 ' ' + this.get('lastName') +
                 ' (' + this.get('login') + ')';
-            name += ' [' + privateName + ']';
         }
-        return name;
+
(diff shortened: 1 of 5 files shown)


brianhelba:
@msmolens I've migrated Study sub-folders on the live database to use this naming, and updated the live site to run this code. I'll wait for your approval to merge, however

web_external/models/UserModel.js
```diff
@@ -1,11 +1,12 @@
 isic.models.UserModel = girder.models.UserModel.extend({
     name: function () {
-        var name;
+        var name = this.get('name');
```
msmolens: This exposes a bug that already existed in some form. The `name` attribute isn't generally defined for users. Right now on the live site I see "undefined [Max Smolens (msmolens)]" in the header user dropdown.

Perhaps this could return null or an empty string if there is no `login` attribute?

brianhelba: Yes, I overlooked the fact that we use `isic.models.UserModel` even in a few places that aren't created by `User.filteredSummary`. I'll fix.

brianhelba:
Fixed.

PR #351, opened 2017-02-28
Prevent ImageViewerWidget from getting a border on focus

brianhelba:


brianhelba:
The GeoJS bugfix is https://github.com/OpenGeoscience/geojs/pull/670

brianhelba:
@msmolens Ok, GeoJS has a new release to fix the bug. This is ready to review.

brianhelba:
@msmolens They didn't exist before GeoJS 0.11, and if we're going to have the capability, it'd be better to document and support it properly as a first class feature.

I'll file a bug to look into it more, but I think the more useful keyboard interactions will be for interacting with annotation-related widgets (which could actually end up conflicting with GeoJS's bindings).

msmolens:
Makes sense, thanks.

PR #356, opened 2017-03-06
Add an API endpoint for featureset deletion

brianhelba:


plugin_tests/featureset_test.py
```diff
@@ -0,0 +1,266 @@
...
+        resp = self.request(
+            path='/featureset/%s' % basicFeatureset['_id'], method='GET')
+        self.assertStatusOk(resp)
+        self.assertDictEqual({
+            '_id': str(basicFeatureset['_id']),
+            '_modelType': 'featureset',
+            'name': basicFeaturesetParams['name'].strip(),
+            'version': float(basicFeaturesetParams['version']),
+            'created':
+                basicFeatureset['created'].replace(tzinfo=pytz.UTC).isoformat(),
```
msmolens: It could be cleaner to apply girder.utility.parseTimestamp` to the `created` field before comparison.

server/provision_utility.py
```diff
@@ -225,4 +166,3 @@ def provisionDatabase():
     _provisionImages()
     _provisionSegmentationGroups()
     _provisionStudies()
-    _provisionDefaultFeatureset()
```
msmolens: Why remove the default "Basic" featureset? I can see that it helps for testing, but is there another reason to change this?

plugin_tests/featureset_test.py
brianhelba: Yes. I'll do so.

server/provision_utility.py
brianhelba: It's not necessary for site operation (as the other magic collections are groups are), so having it in this provisioning script will cause the "Basic" featureset (which likely holds little clinical value) to be re-added to the database after every restart, even if an administrator tries to delete it. IMO, sample / demo data, if we choose to provide it, should be via an op-in mechanism.

msmolens: Sounds good, thanks!

web_external/model.js
```diff
@@ -0,0 +1,17 @@
+isic.Model = girder.Model.extend({
+    urlRoot: function () {
+        return this.resourceName;
+    },
+
+    // "girder.Model.destroy" doesn't trigger many of the proper Backbone events
```
msmolens: What's the impact of this in this plugin?

web_external/model.js
```diff
@@ -0,0 +1,17 @@
...
+    destroy: Backbone.Model.prototype.destroy
+});
+
+Backbone.ajax = function (options) {
```
msmolens: Do we need to override this because we're now calling Backbone's `Model.destroy` directly?

brianhelba:
@msmolens Oops, I pushed a lot of my pending GUI changes that should have been part of branch and PR. I'll back those out now.

The only change from your first review should be an additional commit to fix datetime comparison.

msmolens:
Okay, LGTM.

PR #358, opened 2017-03-07
Add a delete featureset button

### LINKED ISSUE #357: Add a UI for featureset deletion
Add a button to each featureset detail, allowing a study admin user to delete the featureset. The button should permit the operation to fail (if the featureset was in use by a study). The button should ask the user for confirmation before submitting the deletion request.

See https://github.com/ImageMarkup/isic-archive/pull/356 for the featureset deletion API.

brianhelba:
Fixes #357 .

### DIFF
--- a/web_external/Featuresets/FeaturesetView.js
+++ b/web_external/Featuresets/FeaturesetView.js
@@ -1,4 +1,8 @@
 isic.views.FeaturesetView = isic.View.extend({
+    events: {
+        'click .isic-featureset-destroy-button': 'confirmDestroy'
+    },
+
     /**
      * @param {isic.models.FeaturesetModel} settings.model
      */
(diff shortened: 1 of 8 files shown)


web_external/Featuresets/FeaturesetsView.js
```diff
@@ -24,6 +24,11 @@ isic.views.FeaturesetsView = isic.View.extend({
         }, this);
         this.featuresets.fetch();
 
+        // TODO: Use the more general 'update' event, once Backbone is upgraded from Girder
```
msmolens: Does this mean once the version of Backbone used by Girder is upgraded?

web_external/Featuresets/FeaturesetsView.js
```diff
@@ -24,6 +24,11 @@ isic.views.FeaturesetsView = isic.View.extend({
...
+        this.listenTo(this.featuresets, 'remove', function () {
+            this.render();
+        }, this);
```
msmolens: `listenTo` uses the object that it's called on as the context, so the `this` argument shouldn't be here. It's an error that it's written like that above, so that argument can be removed as well.

web_external/models/FeaturesetModel.js
```diff
@@ -10,5 +10,13 @@ isic.models.FeaturesetModel = isic.Model.extend({
 
     creator: function () {
         return new isic.models.UserModel(this.get('creator'));
+    },
+
+    destroy: function (options) {
+        var params = _.clone(options);
```
msmolens: It isn't obvious to me why the clone is necessary. Can you add a comment?

web_external/Featuresets/FeaturesetView.js
```diff
@@ -19,9 +23,45 @@ isic.views.FeaturesetView = isic.View.extend({
...
+                    noText: 'OK'
+                });
+            },
+            error: function (model, resp, options) {
+                girder.confirm({
+                    text: '<h4>Error deleting featureset</h4><br>' + _.escape(resp.responseJSON.message),
+                    escapedHtml: true,
+                    yesClass: 'hidden',
+                    noText: 'OK'
+                });
```
msmolens: I added `isic.showAlertDialog` for informational modal dialogs and it makes sense to use it in place of these `girder.confirm`s.

web_external/Featuresets/FeaturesetsView.js
brianhelba: Yeah. Girder's version is old and doesn't include that event.

web_external/Featuresets/FeaturesetView.js
brianhelba: Cool. Will do.

web_external/Featuresets/FeaturesetsView.js
brianhelba: You're right, but JS just ignores the extra arguments, so I never got a console error. Will fix.

msmolens: Note that in this case it can be simplified further--just use `this.render()` as the third argument.

web_external/models/FeaturesetModel.js
brianhelba: It's good practice for functions not to mutate their parameters unless documented. I want `wait` to always be `true`, regardless of what is passed in. Since `_.extend` will mutate it's first argument, `_.clone` seems like the best way to go.

I'll comment though. Let me know if you see a cleaner way to do this.

web_external/Featuresets/FeaturesetsView.js
brianhelba: Yeah, I actually did that in a place inside the `FeaturesetView`, but didn't see it here.

Also, note the line above this one is also incorrectly passing `this`. That enticed me to do the same here. I'll fix both.

msmolens: Yep, I tried to mention the error in the line above...above, but didn't communicate that well!

brianhelba:
Ok, fixed everything, including some issues with `showAlertDialog` in other places.

PR #361, opened 2017-03-09
Document all View constructors

brianhelba:


web_external/Datasets/DatasetsView.js
```diff
@@ -43,28 +43,27 @@ isic.views.DatasetsView = isic.View.extend({
 
         // Display loading indicator
         if (!this.loaded) {
-            new girder.views.LoadingAnimation({
+            this.loadingAnimation = new girder.views.LoadingAnimation({
                 el: this.$('.isic-listing-loading-animation-container'),
                 parentView: this
             }).render();
+        } else {
```
brianhelba: I split this `else` / `if` on purpose, just because I think it's more understandable.

web_external/Datasets/DatasetView.js
```diff
@@ -2,28 +2,38 @@ isic.views.DatasetView = isic.View.extend({
...
-        }).once('g:fetched', function () {
+        // Display loading indicator
+        this.loadingAnimation = new girder.views.LoadingAnimation({
+            el: this.el,
+            parentView: this
+        }).render();
+
+        this.model.once('g:fetched', function () {
+            this.loadingAnimation.destroy();
+            delete this.loadingAnimation;
```
msmolens: It looks like this caused a bug where the click events don't fire. Perhaps by using `this.el` for the loading animation the events are detached when that view is destroyed.

PR #363, opened 2017-03-14
Fix an issue where non-privileged users couldn't see study annotators

brianhelba:


brianhelba:
I've made this live already.

brianhelba:
Regression testing will catch this sort of issue in the future, provided that we remember to test every endpoint as:
* An anonymous user
* An unprivileged normal user
* A privileged user for that task

msmolens:
I trust it fixes a problem, but it looks like the change only affects the sorting order. Can you explain it a bit more?

brianhelba:
@msmolens The bug was just that since the sorting was being applied to the elements returned after application of `User.filteredSummary`, rather than those returned directly from `Study.getAnnotators`, the key function could not always call `User.obfuscatedName`, since `annotatorUser['login']` may have been removed.

Now, even in the case of an admin user calling this endpoint, this fix removes an extra call to `User.obfuscatedName`.

This is just a straightforward bug that will be caught with better testing (which I'm working on now).

msmolens:
I see that now, thanks for the explanation!

PR #365, opened 2017-03-16
Major refactor of ImagesGallery to use cleaner models and events

brianhelba:


brianhelba:
@ronichoudhury @msmolens I'll talk to you both in person about these changes. Feel free to check out and test the branch, but I would recommend not looking at the code deltas until we have a chance to walk through it together.

There's even more code that can be removed here, but it will require webpack first.

web_external/ImagesGallery/ImagesFilters.js
```diff
@@ -0,0 +1,219 @@
...
+        this._filters = completeFacets.chain()
+            .map(function (completeFacet) {
+                var facetId = completeFacet.id;
+                var facetBins = completeFacet.get('bins');
+
+                var facetFilters = _.chain(facetBins)
+                    .map(function (facetBin) {
+                        var binLabel = facetBin.label;
+                        // Default to all bins included
+                        var binIncluded = true;
```
ronichoudhury: Changing this to `false` should "just work" but it doesn't.

web_external/ImagesGallery/ImagesFilters.js
```diff
@@ -0,0 +1,219 @@
...
+        var excludedBinLabels = _.chain(facetFilters)
+            .pick(function (binIncluded, binLabel) {
+                return binIncluded === false;
+            })
+            .keys()
+            .map(isic.SerializeFilterHelpers._stringToHex)
+            .value();
+        // TODO: Could use "facetId + ' in ' + includedBinLabels"" if most are excluded
+        return '(' +
+            isic.SerializeFilterHelpers._stringToHex(facetId) +
```
ronichoudhury: Putting quote characters around the identifier should allow us to simplify this code (won't need `_stringToHex()` anymore hopefully).

web_external/ImagesGallery/ImagesFilters.js
```diff
@@ -0,0 +1,219 @@
...
+        return '(' + filterExpression + ')';
+    },
+
+    _categoricalFacetFilterExpression: function (facetId, facetFilters) {
+        var excludedBinLabels = _.chain(facetFilters)
+            .pick(function (binIncluded, binLabel) {
+                return binIncluded === false;
+            })
+            .keys()
+            .map(isic.SerializeFilterHelpers._stringToHex)
```
ronichoudhury: This does always have to be "hexified" because the labels come from arbitrary data sources (i.e. we *don't* control the content).

web_external/ImagesGallery/ImagesFilters.js
ronichoudhury: Although... it looks like everything is *de*hexified at the end no matter what, so that may cause problems if this part isn't hexified.

ronichoudhury:
There's a small bug in query.pegjs - the `not` expression should allow whitespace ("`_`") after the opening parenthesis.

web_external/ImagesGallery/ImagesView.js
```diff
@@ -0,0 +1,128 @@
...
+
+            this.images.fetch();
+
+            this.filters.initialize(collection);
+        });
+        this.listenTo(this.filters, 'change', function () {
+            // TODO: Ideally, this should select and cancel only the old set of requests; until
+            // that's fixed, hopefully nobody will click histogram buttons too early in the page
+            // load
+            girder.cancelRestRequests();
```
ronichoudhury: Replace this with a mechanism that saves the two requests made (a few lines below) and only cancel those. This mechanism can break the application because there may be other pending requests that should *not* be canceled by action here.

ronichoudhury:
Push `$rolloverImageFilter` and brethren down into styl file where they are used.

ronichoudhury:
Delete rangeSet.js

web_external/ImagesGallery/ImagesFilters.js
```diff
@@ -0,0 +1,225 @@
...
+        if (fullExpression) {
+            var ast = isic.SerializeFilterHelpers.astParser.parse(fullExpression);
+            ast = isic.SerializeFilterHelpers._dehexify(ast);
+            return isic.SerializeFilterHelpers._specifyAttrTypes(ast);
+        } else {
+            return undefined;
+        }
+    },
+
+    _rangeFacetFilterExpression: function (facetId, facetFilters) {
```
msmolens: - I see errors in this function when toggling the "unknown" groups for "Approximate Age" and "Clinical Size - Longest Diameter (mm)":
    ```
    ImagesFilters.js:100 Uncaught TypeError: Cannot read property '1' of null
    ```

- Toggling "unknown" for "Benign or Malignant" doesn't update the images in the gallery.

- "unknown" handled correctly under "Lesion Diagnosis" either. If I uncheck all the checkboxes, there are still 312 images shown--the exact number of images in the "unknown" category.

- The live site (sometimes) shows more information under "Lesion Diagnosis" (e..g "nevus (3014/12265)").  It looks like this branch always displays only the total.

brianhelba: The first 3 issues here were caused by an incorrect handling of null values. See https://github.com/ImageMarkup/isic-archive/pull/365/commits/c9702d496ea787be3356a4295184eeca56055dd8 for a fix.

brianhelba:
@ronichoudhury PTAL @ https://github.com/ImageMarkup/isic-archive/pull/365/commits/2721949b131e7f8ee16afbe86f221b511df215b1

web_external/ImagesGallery/ImagesFilters.js
brianhelba: Edited my last comment.

web_external/ImagesGallery/ImagesFilters.js
```diff
@@ -101,6 +106,7 @@ _.extend(isic.collections.ImagesFilters.prototype, Backbone.Events, {
                 var highBound = parseFloat(rangeMatches[2]);
                 return [lowBound, highBound];
             })
+            .sortBy(_.identity)
```
ronichoudhury: What effect does this have?

web_external/ImagesGallery/ImagesFilters.js
```diff
@@ -131,10 +137,14 @@ _.extend(isic.collections.ImagesFilters.prototype, Backbone.Events, {
                     ')';
                 return '(' + lowBoundExpression + ' or ' + highBoundExpression + ')';
             })
-            .value()
-            // Combine all expressions
-            .join(' and ');
-        return '(' + filterExpression + ')';
+            .value();
+        if (facetFilters['null'] === false) {
```
ronichoudhury: What's the idea in this conditional block? To filter out all null values if the caller asks for that?

web_external/ImagesGallery/ImagesFilters.js
msmolens: Nice, the "unknown"/nulls work now.

brianhelba:
@ronichoudhury As discussed in our live conversation, note that the changes to `null` handling in serialization of range facet filters have been overwritten by subsequent changes to not use `null` values at all (instead using `'__null__'`. Sorry for the confusion.

ronichoudhury:
No worries - makes sense.

brianhelba:
@msmolens Ok, the last issue should be fixed. This was more complicated, as I realized that the "ZIP" of completeFacetBins and filteredFacetBins wasn't staying updated as filteredFacet was re-fetched (since the nested bins arrays were getting overwritten with new references).

So, I:
* rewrote the `ImagesFacetView` classes to ZIP up completeFacetBins and filteredFacetBins on demand before every rerender
* rewrote the `_zipFacetBins` function to run in linear time
* removed the use of D3 from `ImagesFacetCategoricalView` (so now only the code related to the `ImagesFacetHistogramView` requires D3)
* removed the need for `ImagesFacetCategoricalView._rerenderSelections` (which should be fast, since it needs to be responsive to user clicks) to use the ZIPed bins

This last change then broke the way that `null` bins were being used, as a `null` value cannot be stored as a data attribute on a DOM element. So, to deal with that problem, and the fact that the use of `null` as an object key was also awkward, I rewrote the `ImagesFacetCollection` to always convert bin labels of `null` to `'__null__'` as soon as they're parsed from the REST request, and use `'__null__'` internally everywhere.

In short, (I believe) it should now fully work as intended. Please test and let me know.

msmolens:
@brianhelba I tested again and don't see any problems. Great work!

brianhelba:
@msmolens Cool, thanks. Want to give this a review so that I can merge?

PR #369, opened 2017-03-20
Upgrade Girder to v1.7.0

brianhelba:
To migrate the live server:
* Drop the Job collection, as the log is in a new list-based format, and we have no need for the old job entries.
* Set all superpixel files to have `superpixelFile['exts'] = ['png']`
* Update `girder.local.cfg` to use `=` instead of `:` delimiters
* Disable the "isic-admin" user

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ addons:
             - p7zip-full
 
 before_install:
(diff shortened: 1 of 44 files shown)


brianhelba:
@msmolens PTAL. I haven't fully tested this myself yet, but I wanted to get the changes pushed up at least.

brianhelba:
I've manually tested, and everything seems to work.

PR #371, opened 2017-03-21
Rename "Upload Dataset" to "Create Dataset"

brianhelba:


brianhelba:
Rebased just to deal with a merge conflict.

web_external/layout/layoutHeader.jade
```diff
@@ -79,9 +79,9 @@ nav.navbar.navbar-default.navbar-fixed-top.isic-header-wrapper
                 | Datasets
             if currentUser
               li(role='presentation')
-                a(title='Upload Dataset' href='#dataset/upload')
+                a(title='Contribute Dataset' href='#dataset/create')
                   i.icon-upload
-                  | Upload Dataset
+                  | Contribute Dataset
```
msmolens: I think we're inconsistent in the terminology: "contribute" vs. "create." I'm fine with either here, but maybe it should say "Create Dataset"?

brianhelba: This is intentional. Generally, the consistent and correct term is "create".

For the special case of the global menu, which is more visible across the site, I'm using a different term that will hopefully be more enticing to new users, in the hopes that they'll be more likely to click it and starting the "Request dataset contributor access" process.

PR #372, opened 2017-03-21
Add a delete study button

brianhelba:


brianhelba:
@msmolens PTAL now.

web_external/Studies/StudyView.js
```diff
@@ -48,5 +50,36 @@ isic.views.StudyView = isic.View.extend({
...
+                $('#g-dialog-container').on('hidden.bs.modal', _.bind(this.destroyModel, this));
+            }, this)
+        });
+    },
+
+    destroyModel: function () {
+        this.model.destroy({
+            success: function (model, resp, options) {
+                isic.showAlertDialog({
+                    text: '<h4>Study <b>"' + _.escape(model.name()) + '"</b> deleted</h4>',
```
msmolens: I think we're inconsistent in whether we end text on alert dialogs with a period, but we should review that separately.

PR #380, opened 2017-03-28
Refactor style of gallery paging controls

brianhelba:


web_external/ImagesGallery/Listing/ImagesPagingPane.js
```diff
@@ -53,6 +53,12 @@ isic.views.ImagesPagingPane = Backbone.View.extend({
             staticImageRoot: girder.staticRoot + '/built/plugins/isic_archive/extra/img/'
         }));
 
+        // Tooltips for buttons inside a "btn-group" must be attached to another element, or else
+        // the "btn-group" size will be disrupted when they appear
+        this.$('.btn-group>[data-toggle="tooltip"]').tooltip({
+            trigger: 'hover',
+            container: this.$el
+        });
```
msmolens: Won't this call `.tooltip()` twice on at least some of the tooltip elements, since we already call it below? Perhaps that's harmless, but certainly unnecessary. Can the call below just be removed?

web_external/ImagesGallery/Listing/ImagesPagingPane.js
```diff
@@ -65,52 +71,55 @@ isic.views.ImagesPagingPane = Backbone.View.extend({
...
         // Disable / enable the appropriate paging buttons
-        this.$('#isic-images-seekPrev, #isic-images-seekFirst')
-            .toggleClass('disabled', !this.images.hasPreviousPage());
-
-        this.$('#isic-images-seekNext, #isic-images-seekLast')
-            .toggleClass('disabled', !this.images.hasNextPage());
+        this.$('#isic-images-paging-seek-prev, #isic-images-paging-seek-first')
+            .prop('disabled', !this.images.hasPreviousPage());
+        this.$('#isic-images-paging-seek-next, #isic-images-paging-seek-last')
+            .prop('disabled', !this.images.hasNextPage());
```
msmolens: Girder 2.0 adds the convenience function `girderEnable()` that we should investigate once the plugin is converted.

web_external/ImagesGallery/Listing/ImagesPagingPane.js
```diff
@@ -65,52 +71,55 @@ isic.views.ImagesPagingPane = Backbone.View.extend({
...
                 return (this.images.length / this.filteredFacets.total) * 100 + '%';
             }, this));
 
-        this.$('#isic-images-download-zip').toggleClass(
+        this.$('#isic-images-paging-downloadZip').prop(
             'disabled', this.filteredFacets.total === 0);
+
+        // Any just-disabled buttons will no longer trigger 'mouseleave' or 'focusout' events, so
+        // any still-active tooltips on those buttons must be manually hidden
+        this.$('[data-toggle="tooltip"][disabled]').tooltip('hide');
```
msmolens: Bootstrap docs say "Tooltips on disabled elements require wrapper elements." We should see if adding tooltips to wrapper elements also avoids the need for code like this.

msmolens: Perhaps related, it would be nice for tooltips to show on the disabled buttons. We should see if adding tooltips to wrapper elements allows them to show for the disabled buttons.

web_external/ImagesGallery/Listing/imagesPagingPane.jade
```diff
@@ -1,45 +1,63 @@
...
+        data-placement='bottom'
+        title='Next page'
+      )
+        i.icon-fast-fw
+      button#isic-images-paging-seek-last.btn.btn-sm.btn-default(
+        data-toggle='tooltip'
+        data-placement='bottom'
+        title='Last page'
+      )
+        i.icon-to-end-alt
```
msmolens: My buttons start wrapping to a new line if the text beside it is long enough and my window isn't wide enough. I don't think that used to happen.

web_external/ImagesGallery/imagesColors.styl
```diff
@@ -0,0 +1,9 @@
+$overviewColor = #a8ddb5
+$filteredSetColor = #4eb3d3
+$pageColor = #08589e
+
+$rolloverImageFilter = url(#recolorImageTo00ABFF)
+$selectedImageFilter = url(#recolorImageTo444499)
+$disabledImageFilter = url(#recolorImageToCCCCCC)
```
msmolens: Are all of these still necessary? I believe they were for recoloring SVGs on the fly. IT would be great to remove `isic.shims.recolorImageFilters` altogether if possible.

web_external/ImagesGallery/Listing/imagesPagingPane.jade
msmolens: A clicked button remains depressed/selected/outlined, which doesn't make sense for controls like this. I think an alternative is to not use real buttons, like in the pagination widget, or apply styling hacks.

msmolens: This is an opportunity to choose better icons for these buttons. In particular, I find the double arrows unnecessary. For example, advancing one page is not a "fast-forward" in my mind.

web_external/ImagesGallery/Listing/imagesPagingPane.jade
```diff
@@ -1,45 +1,63 @@
...
+    #isic-images-paging-label
+      span#isic-images-paging-label-loading Loading images...
+      span#isic-images-paging-label-noPagingOrFilters
+        | Showing all&nbsp;
+        span.isic-images-paging-label-page 0
+        |  images in the archive.
```
msmolens: While we're making improvements, I think we could improve this text. Removing "in the archive" simplifies the display without removing information. Text below like "..of the total number of images" could be simplified to something like "Showing x of N images."

brianhelba:
@msmolens PTAL again. I've fixed all that I could from our live discussion.

I put a note in https://github.com/ImageMarkup/isic-archive/issues/382 to revisit the `<button>` conversion, as it causes issues with the event listening and tooltips not being unbound, requiring us to add yet more code to work around. For now, I'd like to get the code as clean as possible, before we build up lots of special cases without a better abstraction mechanism to support it.

brianhelba:
@msmolens Oh, actually, I could also switch the paging icons. Which specific ones (by their `.icon-` name) did you have in mind?

msmolens:
We could try `icon-to-start`, `icon-to-end` for First page and Last page and `icon-play` for Next page. I don't see an icon corresponding to left-pointing "play", but we could apply a `transform: scale(-1, -1)` to the icon element. But maybe that's too complicated, and since no one has mentioned any problems with the current double-arrow controls, I'm okay with leaving them as-is for now.

msmolens:
Also, as a data point, I found an example of an interface that shows a tooltip for a logically disabled element--right on this page:
image

brianhelba:
@msmolens Yeah, let's do the icon thing later. If you can come up with an improved set that you're happy with, feel free to make the PR yourself.

PR #383, opened 2017-03-31
Add buttons to include and exclude all facet bins

### LINKED ISSUE #221: Need a per-histogram button to disable all filters (i.e. all data for that histogram)

brianhelba:
Fixes #221

### DIFF
--- a/web_external/Front/FrontPageView.js
+++ b/web_external/Front/FrontPageView.js
@@ -5,9 +5,7 @@ isic.views.FrontPageView = girder.views.FrontPageView.extend({
 
     render: function () {
         this.$el.html(isic.templates.frontPage({
-            apiRoot: girder.apiRoot,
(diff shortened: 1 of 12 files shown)


brianhelba:
@msmolens PTAL.

msmolens:
This is much needed functionality and works well for me.

One future enhancement, at least for the categorical facets, could be to mimic all or part what http://kayak.com does when searching for flights:
- Provide an "only" link, which appears on hover, to select only a single item.
- Only whenever not all items are selected, provide a "Show all" link. The link is placed next to the section header so it's clear that it's not just another checkbox.

This would optimize the workflow of picking "Select none" followed by selecting an item. Screenshots for reference:
image 
image

brianhelba:
@msmolens Cool, I like the idea. Can you file it as a bug?

Issue #382, opened 2017-03-31
Use enhancements from Girder 2.0


msmolens:
Once the plugin is converted to use Girder 2.0, investigate using the following enhancements:
- [ ] UploadWidget: disallow multiple file selection (Create dataset view, Register metadata view). See https://github.com/girder/girder/blob/7dbab19fc4f37602e6784c1ea3ef7faa34cdf36a/clients/web/src/views/widgets/UploadWidget.js#L103-L104.
- [ ] UploadWidget: use returned file ID to avoid lookups like https://github.com/ImageMarkup/isic-archive/blob/98ce4f001a47ce327acf06c94487caf32f96f1ea/web_external/Datasets/CreateDatasetView.js#L40-L54 (Create dataset view, Register metadata view). See https://github.com/girder/girder/blob/7dbab19fc4f37602e6784c1ea3ef7faa34cdf36a/clients/web/src/views/widgets/UploadWidget.js#L269.
- [ ] Investigate using `girderEnable()` in place of explicitly setting `disabled` attribute. See https://github.com/girder/girder/blob/7dbab19fc4f37602e6784c1ea3ef7faa34cdf36a/clients/web/src/utilities/jquery/girderEnable.js.
- [ ] Verify that displaying modal dialogs in sequence is in fact no longer an issue and remove workarounds like https://github.com/ImageMarkup/isic-archive/blob/3fc6f15126dddb5a6186927806563ce9e655bb85/web_external/Featuresets/FeaturesetView.js#L39-L43. See https://github.com/ImageMarkup/isic-archive/issues/370.

brianhelba:
One we use `girderEnable()`, we should try to switch `button.btn` elements to be `div.btn`, which will prevent unwanted selection styling. Until then, the switching causes issues with `div` not being properly disabled from both tooltip and event listener interactions.

PR #385, opened 2017-04-03
Migrate to Girder 2

brianhelba:
TODO:
- [x] Fix testing
- [x] Include fix from https://github.com/girder/girder/pull/1974 by (temporarily) copy-paste re-implementing `Collection.fetch`
- [x] Fix histogram filter button mouseover styles
- [x] Update readme's Usage section with new build instructions
- [ ] Reconsider how shared style files are included
- [x] See if `publicPath` config option can be removed
- [ ] See if large_image can be updated, removing the need to locally patch it

Future TODO:
- [ ] Rename templates to .pug
- [ ] Use new ES6 language features
- [ ] Use external eslint from #1657
- [ ] Deal with coverage issue #1669
- [ ] Consider testing with circle, #1674
- [ ] Client-side enhancements: https://github.com/ImageMarkup/isic-archive/issues/382


brianhelba:
@msmolens Fixed some additional issues, PTAL.

As far as I can tell, the Swagger problem is just the port numbers getting removed from requests in a development environment. Girder has some significant problems with how it creates absolute URLs to itself, so this may require some upstream patching. However, I believe Swagger will still work in production. We should ensure that worker jobs are also still being created with the correct API URLs, however.

I'm still looking into the issue with the shared style files and `publicPath`.

I'm inclined to defer the `large_image` issue to a later PR. The most recent versions of `large_image` may actually have some additional bugs, so I don't want to risk breaking things further.

msmolens:
@brianhelba I reprovisioned and everything LGTM. I pushed a fix to restore the CSS transitions for collapsing panels. Also, Swagger now works for me in my development environment; I'm not sure what's different from before.

brianhelba:
@msmolens I think this is now ready to go. Let me know what you think.

PR #386, opened 2017-04-04
Add a user invitation feature

brianhelba:
This allows study administrators to pre-create new accounts of behalf of others, which has several benefits:
* the new users will only be required to create a password when first logging in, and will have all other information pre-filled for them
* since the accounts are pre-created, they can be immediately be assigned tasks, allowing the new user to see pending tasks as soon as they first log in
* an invitation email is automatically sent to the new user, with a link to complete the login / registration process


server/api/user.py
```diff
@@ -139,10 +140,76 @@ def acceptTerms(self, params):
...
+        lastName=params['lastName']
+        # TODO: suppress sending emails, in case email verification setting is enabled, as
+        # email verification will also be done when the temporary password is exchanged
+    )
+
+    # TODO: disable user
+
+    token = Token.createToken(
+        newUser, days=validityPeriod,
+        scope=[TokenScope.TEMPORARY_USER_AUTH, TokenScope.EMAIL_VERIFICATION])
```
brianhelba: TODO: When the password is reset, also update the user document to reflect that the email has been verified.

server/api/user.py
```diff
@@ -139,10 +140,76 @@ def acceptTerms(self, params):
...
+
+    html = mail_utils.renderTemplate(
+        'inviteUser.mako',
+        {
+            'newUser': newUser,
+            'inviteUrl': inviteUrl,
+        })
+    mail_utils.sendEmail(
+        to=newUser['email'],
+        subject='ISIC Archive: Dataset Contributor Request',
```
brianhelba: Subject should be changed.

web_external/User/InviteUserView.js
```diff
@@ -0,0 +1,50 @@
...
+                    login: this.$('#isic-user-invite-new-login').val(),
+                    email: this.$('#isic-user-invite-new-email').val(),
+                    firstName: this.$('#isic-user-invite-new-firstname').val(),
+                    lastName: this.$('#isic-user-invite-new-lastname').val()
+                },
+                type: 'POST',
+                error: null
+            })
+                .done(function (resp) {
+                    console.log('success', resp);
```
brianhelba: This should display a confirmation with:
* The user's new obfuscated name
* The URL of the confirmation link
* Anything else that might be useful
We maybe shouldn't render this confirm in a modal, if it's more awkward to copy text out of.

web_external/User/InviteUserView.js
```diff
@@ -0,0 +1,50 @@
...
+            event.preventDefault();
+            // this.$('#isic-user-invite-submit').prop('disabled', true);
+
+            girder.restRequest({
+                path: 'user/invite',
+                data: {
+                    login: this.$('#isic-user-invite-new-login').val(),
+                    email: this.$('#isic-user-invite-new-email').val(),
+                    firstName: this.$('#isic-user-invite-new-firstname').val(),
+                    lastName: this.$('#isic-user-invite-new-lastname').val()
```
brianhelba: This isn't sending the confirm period.

web_external/User/rsvpUserPage.jade
```diff
@@ -0,0 +1,32 @@
...
+        input#isic-user-rsvp-password.input-sm.form-control(type="password", placeholder="Enter new password")
+      .form-group
+        label(for="isic-user-rsvp-password2") Retype new password
+        input#isic-user-rsvp-password2.input-sm.form-control(type="password", placeholder="Retype new password")
+
+  p
+    form#isic-user-rsvp-form(role='form')
+      button#isic-user-rsvp-submit.btn.btn-primary.btn-md(type='submit') Submit
+
+    #g-password-change-error-msg.g-validation-failed-message
```
brianhelba: We can probably remove this line, since errors are now shown in a modal.

brianhelba:
@msmolens Is this ready for review?

msmolens:
@brianhelba Yes, please.

web_external/User/InviteUserView.js
```diff
@@ -2,28 +2,40 @@ isic.views.InviteUserView = isic.View.extend({
     events: {
         'submit #isic-user-invite-form': function (event) {
             event.preventDefault();
-            // this.$('#isic-user-invite-submit').prop('disabled', true);
+            this.$('#isic-user-invite-submit').prop('disabled', true);
```
brianhelba: Let's file a bug to add this same submit-button-disable behavior to our other forms.

web_external/User/InviteUserView.js
```diff
@@ -2,28 +2,40 @@ isic.views.InviteUserView = isic.View.extend({
...
+
+            var data = {
+                login: this.$('#isic-user-invite-new-login').val(),
+                email: this.$('#isic-user-invite-new-email').val(),
+                firstName: this.$('#isic-user-invite-new-firstname').val(),
+                lastName: this.$('#isic-user-invite-new-lastname').val()
+            };
+
+            var validityPeriod = this.$('#isic-user-invite-period').val();
+            if ($.trim(validityPeriod).length > 0) {
```
brianhelba: ES5 has [`String.prototype.Trim`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim). Why not use that?

server/api/user.py
```diff
@@ -159,7 +164,7 @@ def inviteUser(self, params):
     self.requireParams(['login', 'email', 'firstName', 'lastName'], params)
     if 'validityPeriod' in params:
         try:
-            validityPeriod = params['validityPeriod']
+            validityPeriod = float(params['validityPeriod'])
```
brianhelba: Let's `.strip()` this string before coercing it, particularly since the client isn't actually trimming the string before submitting it.

msmolens: The `float` built-in function strips whitespace itself, so I don't think it's necessary.

web_external/User/InviteUserView.js
msmolens: I will update this.

web_external/User/InviteUserView.js
msmolens: I think we do this most places, but I will file a bug to take a comprehensive look.

web_external/User/invitationConfirmationView.js
```diff
@@ -0,0 +1,27 @@
+isic.views.InvitationConfirmationView = isic.View.extend({
```
brianhelba: Our convention is that view files get named with a leading uppercase letter. We also have a semi-convention of always having a (possibly blank) `.styl` file for every view / template group.

I'd be fine with reconsidering both conventions, but they should be followed for now.

web_external/layout/HeaderView.js
```diff
@@ -15,6 +15,7 @@ isic.views.LayoutHeaderView = isic.View.extend({
         this.render();
 
         girder.events.on('g:login', this.render, this);
+        girder.events.on('g:login-changed', this.render, this);
```
brianhelba: FYI, you can bind to multiple events in the same call just by delimiting the event names spaces. Also, we should use `listenTo` in this sort of context, to make it safer when this view is destroyed (though the teardown method should handle it).

However, both things are already fixed in the 2.x branch, so it's not a big deal here.

web_external/User/InviteUserView.js
```diff
@@ -2,28 +2,40 @@ isic.views.InviteUserView = isic.View.extend({
...
-                    console.log('success', resp);
-                })
-                .fail(function (resp) {
-                    isic.showAlertDialog({
-                        text: '<h4>Error sending invite</h4><br>' + _.escape(resp.responseJSON.message),
-                        escapedHtml: true
-                    });
+            .done(function (resp) {
+                resp.newUser.email = data.email;
+                girder.events.trigger('g:navigateTo', isic.views.InvitationConfirmationView, {
```
brianhelba: FYI, I was initially a little skeptical of the need for a whole separate view to show this, and I was hesitant about directly triggering `g:navigateTo` from inside a view like this, instead of the traditional location in a router.

After further consideration though, I think this is indeed the best way. Associating multiple templates with a single view or packing a large amount of conditional code into a single template both seem less maintainable, and we don't want the confirmation to actually live at another route (which would also make it difficult to pass private parameters to). I'd be fine if we did more of this in the future, where appropriate.

server/api/user.py
brianhelba: Yes, you're correct.

web_external/User/InviteUserView.js
brianhelba: I've just poked around a little bit, and most places seem to have it, which I didn't realize. I haven't exhaustively looked at every form, but I don't think a bug report is necessary any more.

web_external/User/invitationConfirmationView.js
brianhelba: Oops, it looks like there is a `.styl` file for this.

Just the filename should be updated.

brianhelba:
Ok, my review pass is done. I haven't tested it fully yet, which I'll do as soon as you finish your  next round of changes. Assuming it checks out then, I'd be happy to merge.

Once you make your next round of changes, feel free to give this PR an acceptable review. I opened it, so I need your review to merge without overrides.

web_external/User/InviteUserView.js
msmolens: Now I'm not so sure this is the best way. The code separate is nice, but this breaks the "Invite New User" menu when on the "Invitation sent" view. Because of that I think one of the other options might be preferable.

web_external/User/InviteUserView.js
```diff
@@ -22,33 +22,44 @@ isic.views.InviteUserView = isic.View.extend({
...
             })
-            .done(function (resp) {
-                resp.newUser.email = data.email;
-                girder.events.trigger('g:navigateTo', isic.views.InvitationConfirmationView, {
-                    newUser: resp.newUser,
-                    inviteUrl: resp.inviteUrl
-                });
-            })
+            .done(_.bind(function (resp) {
+                this.confirmation = _.clone(resp);
```
brianhelba: Why is the `_.clone` necessary? Is the `resp` being shared with something else?

msmolens: I removed this.

brianhelba:
@msmolens Everything LGTM. Incidentally, there is a bug requesting this feature upstream: https://github.com/girder/girder/issues/1143 In the future, we can consider if there are ways to generalize parts or all of this functionality for inclusion upstream.

Travis seems to have broken, but I'm going to merge master back to this as soon as https://github.com/ImageMarkup/isic-archive/pull/396 is merged. That should fix the tests.

PR #387, opened 2017-04-04
Allow deletion of study annotators

brianhelba:


brianhelba:
Just need to fix the alignment of the buttons now.

msmolens:
@brianhelba LGTM. Please take a look at the button changes in the most recent commit.

web_external/Studies/studyPage.jade
```diff
@@ -18,18 +18,24 @@
...
+            title='Add annotator')
             i.icon-plus
     tr
       td.isic-study-users-content
         if study.get('users').length
-          ul
+          table(class=canAdminStudy ? 'table-hover' : undefined)
             each user in study.users().models
-              li= user.name()
+              tr
```
brianhelba: Can we put the `data-user-id=user.id` attribute up on the `tr`, so it's usable by the whole row in the future?

web_external/Studies/StudyView.js
```diff
@@ -15,6 +15,14 @@ isic.views.StudyView = isic.View.extend({
             this.studyAddUserWidget.render();
         },
 
+        'click .isic-study-remove-user-button': function (event) {
+            var target = $(event.currentTarget);
+            var userId = target.parent().data('userId');
```
brianhelba: If you move the `data-user-id`, this line will have to be:
```
var userId = target.closest('data-user-id').data('userId');
```

brianhelba:
@msmolens Great change, this looks good, and the table hover helps even more.

I just think we should make the one tweak (described inline) to where the data is stored.

PR #388, opened 2017-04-05
Refactor image detail API and display

brianhelba:
These are some hopefully-uncontroversial changes, in preparation for the addition of image tags.


brianhelba:
@msmolens PTAL

brianhelba:
@msmolens It must be either Javascript or Jade iteration. I'll go ahead and make it explicit.

brianhelba:
@msmolens Actually, we'll deal with it later. I'll file a bug to investigate.

PR #390, opened 2017-04-05
Add tags field to images, and enable faceted search for them

brianhelba:


brianhelba:
@msmolens PTAL.

I've got another WIP feature to display the tags in the image detail pane, but I'm struggling with the formatting. Let's get this live first.

brianhelba:
Actually, I realize that the filters need to be switched to use `in`, rather than `not in`, for tags. I'll fix this now.

brianhelba:
@msmolens OK, this is ready again. PTAL.

msmolens:
Looks good. There is a difference in behavior between the tags facet and the other facets. When I deselect all the tags, all the images are returned. For other facets this results in no images returned. Is that intentional?

msmolens:
Should tags be shown in the image details?

brianhelba:
@msmolens It's not surprising that it's happening (since we're using inclusion instead of exclusion to filter tags), but there should be a way to fix it. I'll check.

brianhelba:
@msmolens I've fixed the issue with no tag filters allowing all images to show.

I have some code to render tags on the details page, but I'm struggling with the layout. I'd rather get this live, since it's important, then work more (or get your help) on the detail page.

brianhelba:
Just pushed a final commit to rename the null label for no tags to "untagged".

PR #391, opened 2017-04-05
Refactor ImagesFilters, to use specialized FacetFilter members

brianhelba:
This further modularizes the filtering code, permitting the easier incorporation of new facet types (like tags).


brianhelba:
@msmolens PTAL.

This should change nothing externally, but will allow tag filters to be serialized.

PR #392, opened 2017-04-06
Add an image detail display for image tags

brianhelba:


brianhelba:
@msmolens Can you take a look at this, and see if you can get something decent? This version isn't layout properly. I was trying out `.label.label-default`, but I'm open to other ways of styling.

brianhelba:
@msmolens There's no urgency, the important feature was the faceted search for tags.

brianhelba:
@msmolens Is this ready?

msmolens:
@brianhelba Yes. We could update the first commit to remove "WIP" before removing "WIP" from the PR.

web_external/ImagesGallery/Detail/imageDetailsPage.jade
```diff
@@ -86,9 +86,10 @@
...
         td.isic-image-details-table-section-header(colspan=2) Tags
       tr
-        td
+        td(colspan=2)
           each value in image.get('notes').tags
-            span.label.label-default= value
-        td
+            //- Labels separated by a space
+            span.label.label-default.isic-image-details-tag= value
+            | &#x20;
```
brianhelba: Why not use `&nbsp;`? It's a little more idiomatic for HTML than reaching into Unicode.

brianhelba: This obviously isn't a big deal, but IMO, the `|` often makes rendering harder to control. For mixed variable and literal content, I'd just interpolate the variables into a single line, as:
```
span... #{value} &#x20;
```

brianhelba:
@msmolens See my comments. Also, feel free to rebase and squash your commit(s) on top of my WIP one, since mine provides nothing interesting.

web_external/ImagesGallery/Detail/imageDetailsPage.jade
msmolens: - Normally, yes, I'd first reach for `&nbsp;`. But here the intent of the space is to actually break the line, in addition to providing a visual buffer between the labels.
- The space is outside of the span, so it shouldn't be on the same line as the value.

I'll rebase/squash now.

brianhelba:
Looks fine, thanks!

Issue #157, opened 2016-09-02
Add dataset details modal to image gallery


brianhelba:
For each dataset facet bin, an info button should be clickable, which will show details in a modal dialog (or some other convenient rendering).

brianhelba:
We could perhaps render this in a tooltip or some sort of non-modal dialog.

PR #393, opened 2017-04-06
Render the dataset facet as a listing, instead of a histogram

brianhelba:


brianhelba:
These still end up sorted by ID, rather than name, but that's going to be too much work to fix right now.

This layout is easier to read, and will also make addressing https://github.com/ImageMarkup/isic-archive/issues/157 easier.

brianhelba:
@msmolens PTAL.

PR #395, opened 2017-04-13
Ensure image annotators always have read access to that image

brianhelba:


brianhelba:
@msmolens PTAL.

I'd really like to have regression tests here, but we don't have a harness yet to test Image or Study.

custom/static/js/pixelmap.js
```diff
@@ -26,24 +26,8 @@ Pixelmap.prototype.loadImage = function (imageId) {
...
-            }
-            if (c.indexOf(name) === 0) {
-                return c.substring(name.length, c.length);
-            }
-        }
-        return '';
-    }
     $.ajax({
-        url: '/api/v1/item/' + imageId + '/tiles',
-        headers: {'Girder-Token': getCookie('girderToken')}
```
msmolens: Why is the `Girder-Token` cookie no longer necessary? In a local test I get 400 errors loading the images when viewing this page.

brianhelba: Cookie auth should now be allowed on these endpoints. If your test environment has some inconsistencies in the hostname, then maybe it's breaking. This is legacy code, so I'll put the header back in.

msmolens: I tried adding back the header and still get errors, so it's likely a problem with my setup.

msmolens: It looks like my problem is that the angular part is not being rebuilt.

brianhelba:
@msmolens Removed the changed to client-side auth.

brianhelba:
@msmolens Oops, I clobbered your last commit with a rebase force-push. I've now incorporated that into my last commit, though that code in `custom/static/js/dermapp-viewer.js` is for segmentation, which doesn't need the enhanced permission checking.

PR #397, opened 2017-04-13
Show dataset description in Bootstrap popover

### LINKED ISSUE #157: Add dataset details modal to image gallery

msmolens:
On the image gallery, show dataset descriptions in a Bootstrap popover
when hovering over the dataset name.

The facet anchor tags are now contained in divs so that the popover
appears to the right of the text instead of to the right of the entire
panel.

Fixes #157.


msmolens:
If we want to show more information than the description, then this should probably use a modal triggered by a button.

PR #398, opened 2017-04-13
Collapse histogram facets by default

### LINKED ISSUE #394: Histogram facets should default to collapsed state
We now have enough facets that they should start as collapsed, as the pane is currently too long to easily scroll though.

However, histograms facets may fail to render if they are invisible while being rendered (probably due to all the introspective size calculations). If this is the case (and it should be confirmed), then rendering histograms while visible, then hiding the facet once the render completes, would be an acceptable work-around for now.

msmolens:
Collapse histogram facets by default. The initial collapsed state for
any facet--categorical, histogram, etc.--can be specified in the facet
schema.

Fixes #394.


msmolens:
@brianhelba With this change I occasionally see histograms with labels cut off on the bottom, but I can't reliably reproduce it.

msmolens:
@brianhelba I spoke too soon. Toggling a column on a histogram causes all the others to re-render. If a histogram is collapsed when rendered, its labels are cut off.

brianhelba:
@msmolens How about re-rendering a histogram whenever it's expanded?

msmolens:
@brianhelba I had tried that, but the histogram still computes the wrong size unless I delay re-rendering until the box is fully expanded (i.e. `shown.bs.collapse`). One thing to try is caching the computed box height at first render and potentially reusing that value on subsequent renders.

msmolens:
@brianhelba I took at shot at caching some of the properties that cause problems when rendering while the element is hidden.

web_external/ImagesGallery/Facets/imagesFacet.jade
```diff
@@ -1,6 +1,10 @@
...
-    .title= title
-.isic-images-facet-content
+.isic-images-facet-heading(
+  data-toggle='collapse'
+  data-target='##{facetContentId}'
+  )
+  i.isic-images-facet-indicator.icon-down-open
+  span.isic-images-facet-title= title
+.isic-images-facet-content.collapse.in(
+  id='#{facetContentId}'
```
brianhelba: PUG doesn't allow this syntax anymore. It will need to be just `id=facetContentId`. Let's change it now to make migration easier.

web_external/ImagesGallery/Facets/imagesFacet.jade
```diff
@@ -1,6 +1,10 @@
-.attributeSection
-  .header
-    i.toggle.icon-down-open
-    .title= title
-.isic-images-facet-content
+.isic-images-facet-heading(
+  data-toggle='collapse'
+  data-target='##{facetContentId}'
```
brianhelba: This will have to be `data-target='#' + facetContentId` in PUG.

web_external/ImagesGallery/Facets/ImagesFacetView.js
```diff
@@ -14,15 +14,29 @@ isic.views.ImagesFacetView = isic.View.extend({
         this.filter = settings.filter;
 
         this.facetId = this.completeFacet.id;
+        this.facetContentId = this.className + '-' + this.completeFacet.id.replace(/\./g, '-');
```
brianhelba: A minor nit, but let's set this using `this.facetId` instead of `this.completeFacet.id`.

web_external/ImagesGallery/Facets/ImagesFacetView.js
```diff
@@ -14,15 +14,29 @@ isic.views.ImagesFacetView = isic.View.extend({
...
+    _toggleCollapseIndicator: function () {
+        this.$('.isic-images-facet-indicator')
+            .toggleClass('icon-down-open')
+            .toggleClass('icon-right-open');
+    },
 
-            this.$('.isic-images-facet-content').toggle();
+    /**
+     * Apply initial collapse state defined in schema.
+     * If 'collapse' is unspecified, default to true.
```
brianhelba: We have total control over the schema, so do you want to just always define `collpase` within it? Or do you feel like that's too redundant and burdensome?

msmolens:
@brianhelba This is ready for further review.

PR #402, opened 2017-04-21
Return annotation tasks in a random order

brianhelba:


kallooa:
I think I want this done in a specific way. We should discuss.

brianhelba:
As it now stands, the order is random, but deterministic for each study. That is, all raters for a given study will be given the exact same randomized ordering of images (and the ordering is effectively fixed at the time the study's images are added). 

This is in total contrast to the way that segmentations are ordered, where they are nondeterministically totally reshuffled every time the next task is requested.

@kallooa How do you feel about this scheme?

PR #408, opened 2017-05-11
Use new ES6 features

brianhelba:


brianhelba:
I think it's also a good idea to start exclusively using `let` instead of `var` (then use `const` where appropriate too). Most of our code is well written enough to change this via a regex, but there are a few places where I'd want to inspect manually.

Feel free to review this as-is, and I'll consider where the `var` -> `let` change is worthwhile to do in one shot.

ronichoudhury:
Just keep in mind that in some cases you can't switch the ES5 lambdas to ES6 arrow functions - in particular, if you are depending on a framework to call your function with a particular `this` value bound. The only context I know of where this matters is passing functions to d3 selections to set values on the selection by data, but it might crop up elsewhere too (maybe jQuery callbacks but I'm not sure).

brianhelba:
@ronichoudhury Indeed. There are a few places, notably in DOM event handlers and the D3-driven histogram code where we need the particular value of `this` in what is otherwise a lambda. This PR leaves those alone as traditional `function` objects.

brianhelba:
Rebased to fix merge conflicts in:
`web_external/StudyResults/StudyResultsView.js`
`web_external/Tasks/TasksView.js`

web_external/Datasets/ApplyMetadataView.js
```diff
@@ -40,7 +40,7 @@ var MetadataFileCollection = Collection.extend({
     sortDir: SORT_DESC,
 
     parse: function (data) {
-        _.each(data, function (item) {
+        _.each(data, (item) => {
```
msmolens: We could start replacing these with [for...of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of). Some other underscore methods could probably disappear as well (_.map?). This can happen in another pass.

PR #411, opened 2017-05-12
Add user completion info to Manage Annotation Studies

### LINKED ISSUE #405: Add annotation endpoint to check user completion status
@brianhelba I need a way of quickly checking how many annotations a user has submitted for a given study. MSK workstations don't allow use of python to calculate that metric, and given how simple it is, python shouldn't really be necessary. I would like a quick way to get the information I need by going to the web API https://isic-archive.com/api/v1# and entering the studyId and userId.

msmolens:
Fixes #405

### DIFF
--- a/web_external/Studies/studyPage.pug
+++ b/web_external/Studies/studyPage.pug
@@ -28,7 +28,16 @@
       td.isic-study-users-content
         if study.get('users').length
           table(class=study.canAdmin() ? 'table-hover' : undefined)
+            th(colspan=2)
+            th Num. Completed
+            th Percent Completed
             each user in study.users().models
+              - var numCompleted = 'N/A'
+              - var percentCompleted = 'N/A'
+              if study.get('images').length
+                - numCompleted = study.get('userCompletion')[user.id]
+                - percentCompleted = `${(numCompleted / study.get('images').length * 100).toFixed().toString()}%`
+                - numCompleted = `${numCompleted} / ${study.get('images').length}`
               tr(data-user-id=user.id)
                 td.isic-study-user-name= user.name()
                 td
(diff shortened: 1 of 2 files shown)


web_external/Studies/studyPage.pug
```diff
@@ -28,14 +28,25 @@
       td.isic-study-users-content
         if study.get('users').length
           table(class=study.canAdmin() ? 'table-hover' : undefined)
+            th(colspan=2)
+            th Num. Completed
+            th Percent Completed
             each user in study.users().models
+              - var numCompleted = 'N/A'
+              - var percentCompleted = 'N/A'
```
brianhelba: Maybe set `study.get('images').length` to a variable here, since it's used 3 times.

web_external/Studies/studyPage.pug
```diff
@@ -28,14 +28,25 @@
       td.isic-study-users-content
         if study.get('users').length
           table(class=study.canAdmin() ? 'table-hover' : undefined)
+            th(colspan=2)
+            th Num. Completed
+            th Percent Completed
             each user in study.users().models
+              - var numCompleted = 'N/A'
```
brianhelba: We can start using `let` instead of `var` here, since templates are passed through Babel.

brianhelba:
@msmolens Tell me what you think of the changes that I just pushed on top.

msmolens:
@brianhelba I see what you're aiming for, but I find it harder to read when the percent symbols aren't aligned and with the parentheses. Either way is okay with me though.

brianhelba:
@msmolens Ok, take a look at this last try. If it's not too bad, let's just merge to get the feature live.

I have the feeling that we may want to eventually separate the user progress from the add/remove user listing. Showing the current progress state is a different concern from the admin-only task of adding/removing annotators.

msmolens:
@brianhelba LGTM

PR #417, opened 2017-05-21
Remove challenge banner and fix table formatting

brianhelba:


web_external/Front/frontPage.pug
```diff
@@ -5,16 +5,6 @@
       h2 International Skin Imaging Collaboration: Melanoma Project
   .isic-frontpage-body.container
     .row.alert.alert-warning(style='margin-top: 20px; margin-bottom: 0;')
```
msmolens: I think we should remove the 'row' above, otherwise this leaves an empty box.

brianhelba:
@msmolens Oops, fixed.

PR #418, opened 2017-05-23
Girder 2.2

brianhelba:
Addresses some of #382

### DIFF
--- a/web_external/Datasets/CreateDatasetView.js
+++ b/web_external/Datasets/CreateDatasetView.js
@@ -2,7 +2,6 @@ import $ from 'jquery';
 import _ from 'underscore';
 
 import FolderModel from 'girder/models/FolderModel';
-import ItemCollection from 'girder/collections/ItemCollection';
 import UploadWidget from 'girder/views/widgets/UploadWidget';
 import {getCurrentUser} from 'girder/auth';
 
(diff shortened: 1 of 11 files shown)


brianhelba:
@msmolens I need to do more testing on this myself, but feel free to start looking.

brianhelba:
The commit for `Displaying modal dialogs in sequence now works natively` doesn't seem to actually work for me.

msmolens:
@brianhelba Yes, I still see the dialog issue in some cases, but not always, so there's likely a race condition. I agree that we should drop that commit for now.

web_external/Datasets/CreateDatasetView.js
```diff
@@ -179,8 +160,8 @@ const CreateDatasetView = View.extend({
         }
     },
 
-    uploadFinished: function (files) {
-        this.uploadedZipFiles = _.pluck(files.files, 'name');
+    uploadFinished: function (info) {
+        this.uploadedFiles = info.files;
```
msmolens: I'm seeing behavior where the File object in `info.files` loses its `id` attribute. I think it's safer to explicitly save the attributes we need, like:

````
this.uploadedFiles = _.map(info.files, (file) => ({id: file.id, name: file.name}));
````

File: web_external/models/UserModel.js [hunk 2]
```diff
@@ -33,39 +34,53 @@ UserModel.prototype.canAcceptTerms = function () {
...
+        } else {
+            // This should not fail
+            deferred.reject(resp);
         }
-        // This should not fail
+    })
+    .fail((resp) => {
+        deferred.reject(resp);
     });
+    return deferred.promise();
```
msmolens: I think we should be able to avoid this pattern of creating and resolving/rejecting a new Deferred. Something like:
````
    return restRequest({
        path: 'user/acceptTerms',
        type: 'POST'
    })
    .done((resp) => {
        if (_.has(resp, 'extra') && resp.extra === 'hasPermission') {
            // Directly update user permissions
            this.get('permissions').acceptTerms = true;
            this.trigger('change:permissions');
    })
````

File: web_external/models/UserModel.js [hunk 3]
```diff
@@ -33,39 +34,53 @@ UserModel.prototype.canAcceptTerms = function () {
...
+            deferred.resolve(resp);
         } else {
-            failureCallback(resp);
+            deferred.reject(resp);
         }
+    })
+    .fail((resp) => {
+        deferred.reject(resp);
     });
+    return deferred.promise();
```
msmolens: Here too. Maybe:
````
    return restRequest({
        path: 'user/requestCreateDatasetPermission',
        type: 'POST'
    })
    .then((resp) => {
        if (_.has(resp, 'extra') && resp.extra === 'hasPermission') {
            // Directly update user permissions
            this.get('permissions').createDataset = true;
            this.trigger('change:permissions');
        } else {
            return $.Deferred().reject(resp);
        }
    })
````

brianhelba: Totally agreed. jQuery 2 is a little weird in how it handles `.then`, so I wanted to be safe, but just returning a thenable should still work.

web_external/models/UserModel.js
brianhelba: I mostly agree. We'll still have to create a new Deferred here, but we can do it within the `.then`. Ideally, we should just be able to use `return` and `throw` within `.then`, but jQuery 2 doesn't support that yet.

web_external/Datasets/CreateDatasetView.js
brianhelba: Sounds good. It's probably not a good idea to trust that strange system-created objects will be immutable.

brianhelba:
@msmolens Really good feedback, thanks. I'l fix these all.

brianhelba:
@msmolens Ok, I've dropped the `Displaying modal dialogs in sequence now works natively` commit from the middle and added two new commits at the end to address your feedback.

PTAL.

brianhelba:
@msmolens Full disclosure, I haven't had a chance to test these last two changes. I'll do so later if you don't get to it first.

PR #429, opened 2017-08-22
Upgrade to Girder 2.3.0

brianhelba:


msmolens:
Nice work, there's a lot here, and it's great to update to the latest Girder.

On the backend I too see the 'originalId' error during upload.

On the frontend the global font is different on my Linux system from the live site. This causes an additional problem on the Study Results list of images where the thumbnails appear below the image name instead of beside it.

Additionally, does the following comment apply now that it uses `girderEnable`?: https://github.com/ImageMarkup/isic-archive/issues/382#issuecomment-290788450

msmolens:
On a fresh deployment with the latest branch, the isic_archive and large_image plugins don't load correctly, so no data is returned when browsing to http://localhost:8080. The Girder interface is available at /girder.

brianhelba:
The provisioning now works reliably on Vagrant instances and (with some minor modification) was able to provision a new `stage.kitware.com`.

brianhelba:
The upload processing issues were fixed by upstream commits to `girder_worker`, which this branch now pulls down.

brianhelba:
We can handle the rest of #382 later.

Merging.

PR #430, opened 2017-08-24
Add script to manage snapshots of volumes on an AWS EC2 instance

msmolens:


danlamanna:
For metabolomics I ended up using this repository for managing snapshots: https://github.com/skymill/automated-ebs-snapshots. It's not incredibly active, but I think that's due to the stable nature of the project rather than any lack of attention.

I wanted daily backups retained for 7 days and what it required was running this:
```
automated-ebs-snapshots --watch $volume_id --interval daily --retention 7
```

And then adding an entry to crontab (every 5 minutes) that looked like this:
```
automated-ebs-snapshots --run
```

Ansible to set this up (which is very short) is in the metabolomics deployment folder on the internal gitlab.

PR #431, opened 2017-09-14
Upgrade VIPS and Vagrant OS

brianhelba:
* Upgrade VIPS
* Activate virtualenv on Vagrant login
* Use Ubuntu 16.04 for Vagrant


brianhelba:
@msmolens This should now be working. I just tested a fresh provision of Vagrant (`vagrant destroy -f && vagrant up`), and I was able to:
* load the web app in my host browser
* login with the provision admin user
* upload a dataset
* see that the pre-processing (superpixels and TIFF) completed

all without having to `vagrant ssh` to fix anything.

If you want to test this yourself, I can wait, but otherwise, I'm happy with merging it now.

PR #438, opened 2017-10-06
Metadata: Show warnings for unrecognized fields

msmolens:
When applying metadata to a dataset, show warnings for fields that
weren't recognized and will be added to unstructured metadata.

### DIFF
--- a/server/models/dataset_helpers/image_metadata.py
+++ b/server/models/dataset_helpers/image_metadata.py
@@ -472,20 +472,21 @@ def addImageClinicalMetadata(image, data):
     Returns a tuple of:
     - List of descriptive errors with the metadata. An empty list indicates that
       there are no errors.
-    - Dictionary view object of unrecognized field names. To avoid erroneous
-      entries, this is populated only when there are no errors.
+    - List of warnings about the metadata. To avoid erroneous entries, this is
+      populated only when there are no errors.
 
     :param image: The image.
     :type image: dict
     :param data: The image metadata.
     :type data: dict
-    :return: Tuple of (errors with the metadata, unrecognized field names)
-    :rtype: tuple(list of strings, dictionary view object or None)
+    :return: Tuple of (errors, warnings)
+    :rtype: tuple(list of strings, list of strings)
     """
     # Operate on copy of image metadata
     data = copy.deepcopy(data)
 
     errors = []
+    warnings = []
 
     for parser in [
         AgeFieldParser,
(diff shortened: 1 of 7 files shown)


server/models/dataset_helpers/image_metadata.py
```diff
@@ -515,4 +518,7 @@ def addImageClinicalMetadata(image, data):
     # Add remaining data as unstructured metadata
     image['meta']['unstructured'].update(data)
 
-    return errors
+    # Report unrecognized fields when there are no errors
+    unrecognizedFields = six.viewkeys(data) if not errors else None
```
brianhelba: I think a `list` would be simpler here, and adequate for now. It can just be empty if there are no `unrecognizedFields`.

This does bring up the point that we're mutating the `data` parameter in-place. I'm not sure if this is a good design, as we could make a copy of `data` before calling each of the `parser.run` on it. @msmolens What do you think?

web_external/Datasets/ApplyMetadataView.js
```diff
@@ -51,23 +51,25 @@ const MetadataFileCollection = Collection.extend({
...
-// Collection of metadata errors. The list of items in the collection is
-// meaningful only when initialized() is true.
-const MetadataErrorCollection = Backbone.Collection.extend({
-    model: MetadataErrorModel,
+// Collection of metadata results, i.e. errors or warnings. The 'field' property
+// should be set to the name of the field to parse in the response from the server.
+// The list of items in the collection is meaningful only when initialized() is true.
+const MetadataResultModelCollection = Backbone.Collection.extend({
+    model: MetadataResultModel,
+    field: null,
```
brianhelba: You can add an `initialize` method to the `MetadataResultModelCollection` to take `field`.

Just call `Backbone.Collection.prototype.apply(this, arguments);` as the first line, to make it transparent.

server/models/dataset_helpers/image_metadata.py
```diff
@@ -515,4 +518,7 @@ def addImageClinicalMetadata(image, data):
...
+
+    return errors, unrecognizedFields
```
brianhelba: Ideally, this would return a tuple of general `(errors, warnings)`, but this wouldn't work with the logic to only capture the first row's "warnings". I can't think of a good solution right now, though.

msmolens: I changed the logic at the model layer and changed this function to return `(errors, warnings)`.

server/models/dataset_helpers/image_metadata.py
msmolens: Now this function operates on a copy of `data`.

web_external/Datasets/ApplyMetadataView.js
msmolens: I added an `initialize` method to accept the `field`.

msmolens: The default Backbone.Collection `initialize()` is documented as empty, so I think it's common to omit the prototype call in this case.

brianhelba: Agreed.

PR #451, opened 2017-10-12
Allow extra images in metadata CSV file

### LINKED ISSUE #448: Metadata validation should emit a warning on unmatched files
CSV files should be allowed to contain superfluous rows, which do not match any existing images names / original filenames. A warning should be emitted for each of these rows.

msmolens:
When validating metadata, allow images to be defined in the CSV file that aren't in the database. Report these images as warnings. Previously these cases were reported as errors.

Fixes #448

### DIFF
--- a/server/models/dataset.py
+++ b/server/models/dataset.py
@@ -419,7 +419,7 @@ class Dataset(FolderModel):
             isicIdField = None
         if (not originalNameField) and (not isicIdField):
             raise FileMetadataException(
-                'no "filename" or "isic_id" field found in CSV')
+                'no \'filename\' or \'isic_id\' field found in CSV')
         return originalNameField, isicIdField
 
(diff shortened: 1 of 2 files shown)


server/models/dataset.py
```diff
@@ -390,7 +390,9 @@ def applyMetadata(self, dataset, metadataFile, save):
 
                     # Add updated image to list of images to potentially save
                     images.append(image)
-                except RowMetadataException as e:
+                except RowMetadataWarning as e:
```
brianhelba: It's ok to use exception handling to get capture warnings in this case, since a failure to match in `_getImageForMetadataCsvRow` is still "fatal" and will abort processing of the whole row. In the future, I anticipate that we might have non-fatal warnings for in rows, which will need to be captured differently.

PR #453, opened 2017-10-12
Update inter-field validation rules

### LINKED ISSUE #447: Update inter-field validation rules
Update the inter-field validation rules to use the logic (here in pseudo-code):
```python
if diagnosis == 'melanoma'
    if benign_malignant == None
        benign_malignant = 'malignant'
    assert benign_malignant == 'malignant'

if diagnosis in ['basal cell carcinoma', 'squamous cell carcinoma']
    if benign_malignant in ['benign', 'indeterminate/benign']
        emit warning "${diagnosis} is typically not ${benign_malignant}"

if diagnosis == 'nevus'
    if benign_malignant == None and diagnosis_confirm_type not in [None, 'histopathology']
        benign_malignant = 'benign'
    assert benign_malignant in ['benign', 'indeterminate/benign', 'indeterminate']

if benign_malignant in ['malignant', 'indeterminate', 'indeterminate/benign', 'indeterminate/malignant']
    assert diagnosis_confirm_type == 'histopathology'
```
Note that we can now sometimes set unknown fields, based on other known field values.

msmolens:
Fixes #447

### DIFF
--- a/server/models/dataset_helpers/image_metadata.py
+++ b/server/models/dataset_helpers/image_metadata.py
@@ -473,39 +473,77 @@ class NevusTypeFieldParser(FieldParser):
 #                 sorted(allowed_diagnoses))
 
 
-def _validateMetadata(clinical):
+def _populateMetadata(clinical):
     """
-    Validate metadata with respect to consistency between fields. Fields with
-    unknown values always validate. An InconsistentValuesException is raised
-    if a value violates one of the pre-defined rules.
+    Populate empty metadata fields that can be determined based on other fields.
     """
     diagnosis = clinical.get('diagnosis')
     benignMalignant = clinical.get('benign_malignant')
     diagnosisConfirmType = clinical.get('diagnosis_confirm_type')
 
-    if diagnosis in [
-        'basal cell carcinoma',
-        'melanoma',
-        'squamous cell carcinoma',
-    ] and benignMalignant is not None and benignMalignant != 'malignant':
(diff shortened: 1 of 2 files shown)


server/models/dataset_helpers/image_metadata.py
```diff
@@ -475,37 +486,55 @@ def load(cls, value, clinical, private):
...
-        raise InconsistentValuesException(
-            names=[DiagnosisFieldParser.name, BenignMalignantFieldParser.name],
-            values=[diagnosis, benignMalignant])
-
-    if benignMalignant in [
+    ] and clinical.get('benign_malignant') in [
+        'benign',
+        'indeterminate/benign',
+    ]:
+        raise AtypicalBenignMalignantException(
```
brianhelba: We don't want to abort processing here, as the remaining rules may still with an error.

server/models/dataset_helpers/image_metadata.py
```diff
@@ -70,6 +70,17 @@ def __init__(self, names, values):
         self.values = values
 
 
+class AtypicalBenignMalignantException(MetadataFieldException):
```
brianhelba: I think we can name thing something slightly more general, like `AtypicalValueException`, provided that you continue to use exceptions as the mechanism for conveying these warnings.

msmolens:
Thanks for the suggestions. I split validation into the three steps described above. Note: I also rebased to test with other recent metadata changes.

PR #459, opened 2017-10-19
Add an option to download image metadata in ZIP

### LINKED ISSUE #449: Include metadata in images ZIP
Add a new parameter `include=` with values `all` `images` `metadata`.
Add a button to the ZIPfile download, with a dropdown to select either of these endpoints.

brianhelba:
Fixes #449

### DIFF
--- a/server/api/image.py
+++ b/server/api/image.py
@@ -181,7 +181,8 @@ class ImageResource(IsicResource):
 
         include = params.get('include', 'all')
         if include not in {'all', 'images', 'metadata'}:
-            raise RestException('Param "include" must be one of: "all", "images", "metadata"')
+            raise ValidationException(
+                'Param "include" must be one of: "all", "images", "metadata"', 'include')
 
         if 'filter' in params:
             query = self._parseFilter(params['filter'])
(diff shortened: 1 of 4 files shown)


server/api/image.py
```diff
@@ -115,10 +177,12 @@ def find(self, params):
     @access.cookie
     @access.public
     def downloadMultiple(self, params):
-        Dataset = self.model('dataset', 'isic_archive')
-        File = self.model('file')
         Image = self.model('image', 'isic_archive')
 
+        include = params.get('include', 'all')
+        if include not in {'all', 'images', 'metadata'}:
+            raise RestException('Param "include" must be one of: "all", "images", "metadata"')
```
msmolens: Often we use (abuse?) `ValidationException` for invalid parameter values. Here's an example in this file: https://github.com/ImageMarkup/isic-archive/blob/c9c2e415d62eadcbcfcd368c067407b4a5112412/server/api/image.py#L348-L349. 
This could be written like:
```python
raise ValidationException('Value must be one of ...', 'include')
```

brianhelba: Yeah, these rules are not really clear anywhere in Girder, but I agree that you suggestion is more consistent with other places.

brianhelba:
@msmolens Fixed. PTAL.

PR #461, opened 2017-10-19
Add a button to zoom directly into images in gallery

brianhelba:
image

### DIFF
--- a/web_external/ImagesGallery/Listing/ImageWall.js
+++ b/web_external/ImagesGallery/Listing/ImageWall.js
@@ -15,11 +15,7 @@ const ImageWall = View.extend({
             let clickedImage = this.images.get(imageId);
 
             if (event.shiftKey) {
-                new ImageFullscreenWidget({ // eslint-disable-line no-new
-                    el: $('#g-dialog-container'),
-                    model: clickedImage,
-                    parentView: this
-                }).render();
+                this._showImageZoomModal(clickedImage);
             } else {
                 clickedImage.toggleSelected();
             }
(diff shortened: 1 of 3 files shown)


web_external/ImagesGallery/Listing/ImageWall.js
```diff
@@ -23,6 +23,19 @@ const ImageWall = View.extend({
...
+
+        'click .isic-images-imageWall-zoom': function (event) {
+            let imageId = $(event.currentTarget).parent().parent().data('imageId');
+            let clickedImage = this.images.get(imageId);
+
+            new ImageFullscreenWidget({ // eslint-disable-line no-new
+                el: $('#g-dialog-container'),
+                model: clickedImage,
+                parentView: this
+            }).render();
```
msmolens: Nit: This `new ImageFullscreenWidget({ ...` could be refactored, since the other `click` handler performs the same operation.

web_external/ImagesGallery/Listing/ImageWall.js
```diff
@@ -23,6 +23,19 @@ const ImageWall = View.extend({
...
+
+            event.stopPropagation();
```
msmolens: I usually see this called first thing in the event handler, but it shouldn't matter when it's called.

brianhelba:
@msmolens Fixed. PTAL.

web_external/ImagesGallery/Listing/ImageWall.js
brianhelba: Yes, but the consistent style is a good point.

web_external/ImagesGallery/Listing/ImageWall.js
brianhelba: I was sort of split on whether to do this, but it's not a bad idea.

PR #462, opened 2017-10-19
Reformat image gallery metadata pane

brianhelba:


web_external/ImagesGallery/Detail/imageDetailsPage.pug
```diff
@@ -27,8 +27,55 @@
 
   table.table.table-condensed.isic-image-details-table
     tbody
+      - const meta = image.get('meta');
+      if !_.isEmpty(meta.clinical)
+        tr
+          td.isic-image-details-table-section-header(colspan=2) CLINICAL INFORMATION
```
msmolens: The ALL-CAPS here is not to my taste, but if it meets a design requirement I'm okay with it. This could be done in CSS with `text-transform: capitalize`.

brianhelba: I agree about the aesthetics, but let's consider it a design requirement for now. We can always iterate once the overall UI is closer to something our users are really happy with.

I'll make the CSS change though.

brianhelba:
@msmolens Updated. PTAL.

PR #468, opened 2017-10-23
Add inter-field validation for melanocytic field

### LINKED ISSUE #458: Add an inter-field validation rule for `melanocytic`
Implement logic equivalent to:
```
if diagnosis in [None, 'other']:
    pass
elif diagnosis in [
    'AIMP',
    'melanoma',
    'melanoma metastasis',
    'nevus',
    'nevus spilus',
    'atypical melanocytic proliferation',
    'solar lentigo',
    'lentigo simplex',
    'lentigo NOS'
]:
    if melanocytic is None:
        melanocytic = True
    assert melanocytic == True
else:
    if melanocytic is None:
        melanocytic = False
    assert melanocytic == Flase
```

msmolens:
Set and validate melanocytic field based on diagnosis.

Fixes #458

### DIFF
--- a/server/models/dataset_helpers/image_metadata.py
+++ b/server/models/dataset_helpers/image_metadata.py
@@ -21,6 +21,19 @@ import copy
 import six
 
 
+_melanocyticDiagnoses = [
+    'AIMP',
+    'melanoma',
+    'melanoma metastasis',
+    'nevus',
+    'nevus spilus',
+    'atypical melanocytic proliferation',
+    'solar lentigo',
+    'lentigo simplex',
+    'lentigo NOS',
+]
+
+
 class MetadataFieldException(Exception):
     """Base class for exceptions raised while parsing metadata fields."""
     pass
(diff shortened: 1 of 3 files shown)


server/models/dataset_helpers/image_metadata.py
```diff
@@ -21,6 +21,19 @@
 import six
 
 
+_melanocyticDiagnoses = [
```
brianhelba: Can you put this as a class object under `MelanocyticFieldParser`? As we get more of these, it will be messy to have lots of global objects floating around.

server/models/dataset_helpers/image_metadata.py
```diff
@@ -513,6 +535,18 @@ def _checkMetadataErrors(clinical):
                 names=[DiagnosisFieldParser.name, BenignMalignantFieldParser.name],
                 values=[diagnosis, benignMalignant])
 
+    # Set melanocytic field based on diagnosis
+    if diagnosis in _melanocyticDiagnoses:
+        if not melanocytic:
```
brianhelba: Since `melanocytic` could also be `None`, I think it's cleaner to explicitly check whether it's `True` or `False` here and below.

server/models/dataset_helpers/image_metadata.py
```diff
@@ -513,6 +535,18 @@ def _checkMetadataErrors(clinical):
...
+            raise InconsistentValuesException(
+                names=[DiagnosisFieldParser.name, MelanocyticFieldParser.name],
+                values=[diagnosis, melanocytic])
+    elif diagnosis is not None and diagnosis != 'other':
+        if melanocytic:
```
brianhelba: Here too.

server/models/dataset_helpers/image_metadata.py
msmolens: The validation process has already populated the field with `True` or `False` at this point, right?

brianhelba: Yes, but I think it's more robust to not have to depend on that assumption in later parts of the code.

PR #469, opened 2017-10-23
Require units to be specified for clin_size_long_diam_mm field

### LINKED ISSUE #464: Require units to be specified for clin_size_long_diam_mm
For `clin_size_long_diam_mm`, rather than accepting a simple numeric value, we should always require units to be specified (as a string suffix). `um`, `mm`, `cm` are acceptable, and the validation logic should convert anything to `mm` in the database.

msmolens:
The clin_size_long_diam_mm now requires units in um, mm, or cm to be
specified. The value is converted to mm when stored in the database.

Fixes #464

### DIFF
--- a/server/models/dataset_helpers/image_metadata.py
+++ b/server/models/dataset_helpers/image_metadata.py
@@ -18,6 +18,9 @@
 ###############################################################################
 
 import copy
+import functools
+import math
+import re
 import six
 
 
@@ -137,6 +140,8 @@ class FieldParser(object):
     def _coerceFloat(cls, value):
         try:
             value = float(value)
+            if math.isinf(value) or math.isnan(value):
+                raise ValueError
         except ValueError:
             raise BadFieldTypeException(name=cls.name, fieldType='float', value=value)
         return value
@@ -260,8 +265,12 @@ class PersonalHxMmFieldParser(HxMmFieldParser):
 
 
 class ClinicalSizeFieldParser(FieldParser):
+    """
+    Parse clinical size field. Expects units to be specified (um, mm, or cm).
+    """
     name = 'clin_size_long_diam_mm'
     allowedFields = {'clin_size_long_diam_mm'}
+    _formatRegex = re.compile(r'(.+)(um|mm|cm)$')
 
     @classmethod
     def transform(cls, value):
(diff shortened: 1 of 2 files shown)


server/models/dataset_helpers/image_metadata.py
```diff
@@ -260,8 +261,12 @@ class PersonalHxMmFieldParser(HxMmFieldParser):
 
 
 class ClinicalSizeFieldParser(FieldParser):
+    """
+    Parse clinical size field. Expects units to be specified (um, mm, or cm).
+    """
     name = 'clin_size_long_diam_mm'
     allowedFields = {'clin_size_long_diam_mm'}
+    _formatRegex = re.compile(r'(.+)\s?(um|mm|cm)$')
```
brianhelba: There's not really a point to this single `\s?` as-is, since other whitespace characters can be matched by `.+` and (multiple) whitespaces can be passed into the [`float` function](https://docs.python.org/2/library/functions.html#float). Either more strict number/whitespace matching should be done in this regex, or it should all be done by `_coerceFloat`. IMO, we should do the latter, since that's more general/shared code.

server/models/dataset_helpers/image_metadata.py
```diff
@@ -271,7 +276,19 @@ def transform(cls, value):
...
+                result = re.match(cls._formatRegex, value)
+                if not result:
+                    raise BadFieldTypeException(
+                        name=cls.name, fieldType='float with units (um, mm, or cm)', value=value)
+                value, units = result.groups()
                 value = cls._coerceFloat(value)
+
+                # Convert to mm
+                if units == 'um':
+                    value /= 100
```
brianhelba: ”um" is micrometers, and is 10^-6. "mm" is 10^-3. So the conversation factor is 1000.0.

brianhelba: Let's use explicit decimals for both of these factors. In Python2 an `int` divided by an `int` is a truncated `int`. Even though we expect a `float` from `_coerceFloat`, there's no reason to have another implicit assumption.

msmolens:
@brianhelba PTAL

server/models/dataset_helpers/image_metadata.py
```diff
@@ -137,6 +140,8 @@ def _coerceInt(cls, value):
     def _coerceFloat(cls, value):
         try:
             value = float(value)
+            if math.isinf(value) or math.isnan(value):
+                raise ValueError
```
brianhelba: Typically, I would expect to see `raise ValueError()`, but it looks like this form will automatically instantiate the passed class: https://docs.python.org/2/reference/simple_stmts.html#raise , and we use this syntax in other places too.

PR #476, opened 2017-10-24
Fix temporary directory lifetime when extracting zip using fallback

### LINKED ISSUE #470: Ensure that ZipFileOpener always has a TempDir open
Currently, the `ZipFileOpener._fallbackUnzip` path is prematurely closing the `TempDir`, before the iterator has been exhausted.

A possible solution is to make the `TempDir` an instance variable of `ZipFileOpener`, and create and destroy the `TempDir` explicitly in `ZipFileOpener__enter__` and `ZipFileOpener__exit__`.

msmolens:
Fix a bug where when using the 7-Zip fallback where the temporary directory into which the zip file is extracted is deleted prematurely.

Fixes #470

### DIFF
--- a/server/upload.py
+++ b/server/upload.py
@@ -53,13 +53,18 @@ class ZipFileOpener(object):
         # TODO: check for "7z" command
 
     def __enter__(self):
+        # Create temporary directory
+        self.tempDirManager = TempDir()
+        self.tempDir = self.tempDirManager.__enter__()
+
         try:
             return self._defaultUnzip()
         except (zipfile.BadZipfile, NotImplementedError):
             return self._fallbackUnzip()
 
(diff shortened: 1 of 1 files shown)


server/upload.py
```diff
@@ -53,15 +53,20 @@ def __init__(self, zip_file_path):
         # TODO: check for "7z" command
 
     def __enter__(self):
+        # Create temporary directory
+        self.tempDirManager = TempDir()
+        tempDir = self.tempDirManager.__enter__()
```
brianhelba: Is there a reason that you didn't make `tempDir` also an instance member, so we don't have to pass it through to every function?

msmolens: I guess that's better here. I'll update this branch.

PR #480, opened 2017-10-30
 Add facets for new metadata types

### LINKED ISSUE #456: Expose new structured metadata fields to faceted search
`dermoscopic_type` / `image_type` / `melanocytic`, etc. need to be exposed to faceted search.

brianhelba:
Fixes #456


brianhelba:
@msmolens As of this PR, we should be all caught up in terms of syncing our actual database + the validation logic + the exposed facets. So, from now on, when we add new validated metadata types, let's also make the change to expose the field as a searchable facet.

PR #484, opened 2017-11-01
Use natural sorting for metadata warnings

msmolens:
Use natural sorting for metadata warnings instead of lexicographic
sorting. Because the warnings include CSV row numbers, natural sorting
ensures that the warnings appear in row order.

A dependency is added on a module that provides a natural sorting
algorithm [1].

[1] https://github.com/SethMMorton/natsort

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 geojson>=1.3.2
 python-dateutil>=2.6.0
+natsort>=5.0.0
 numpy>=1.10.2
 
 Pillow==3.4.2
--- a/server/models/dataset.py
+++ b/server/models/dataset.py
@@ -22,6 +22,8 @@ import datetime
 import itertools
 import os
 
+from natsort import natsorted
+
 from girder.constants import AccessType
 from girder.models.folder import Folder as FolderModel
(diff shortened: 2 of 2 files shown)


kallooa:
If this will take more than 10 minutes to implement it doesn't really sound like something worth doing. If it is a one or two line change then go for it

server/models/dataset.py
```diff
@@ -28,6 +28,7 @@
 from girder.models.notification import ProgressState
 from girder.utility import assetstore_utilities, mail_utils
 from girder.utility.progress import ProgressContext
+from natsort import natsorted
```
brianhelba: Since this is a third-party package dependency, can you put it in its own group between the builtin module imports and the Girder module imports?

requirements.txt
```diff
@@ -1,5 +1,6 @@
 geojson>=1.3.2
 python-dateutil>=2.6.0
+natsort==5.1.0
```
brianhelba: Can this be a more abstract version? I'd like to move away from using locked down requirements for packages (except in a lockfile, which we don't actually have).

PR #485, opened 2017-11-01
Require metadata CSV to be UTF-8 encoded

### LINKED ISSUE #473: Saving metadata with Unicode characters results in database errors
This appears only in the final saving of metadata, not in the read-only validation.

This example came from an unstructured metadata field containing a Unicode value:
```
Traceback (most recent call last):
  File "/home/ubuntu/girder/girder/api/rest.py", line 567, in endpointDecorator
    val = fun(self, args, kwargs)
  File "/home/ubuntu/girder/girder/api/rest.py", line 1137, in POST
    return self.handleRoute(method, path, params)
  File "/home/ubuntu/girder/girder/api/rest.py", line 890, in handleRoute
    val = handler(**kwargs)
  File "/home/ubuntu/girder/girder/api/access.py", line 63, in wrapped
    return fun(*args, **kwargs)
  File "/home/ubuntu/girder/girder/api/rest.py", line 389, in wrapped
    return fun(*args, **kwargs)
  File "/home/ubuntu/girder/girder/api/rest.py", line 389, in wrapped
    return fun(*args, **kwargs)
  File "/home/ubuntu/girder/plugins/isic_archive/server/api/dataset.py", line 329, in applyMetadata
    dataset=dataset, metadataFile=metadataFile, save=save)
  File "/home/ubuntu/girder/plugins/isic_archive/server/models/dataset.py", line 405, in applyMetadata
    Image.save(image)
  File "/home/ubuntu/girder/girder/models/model_base.py", line 412, in save
    {'_id': document['_id']}, document, True)
  File "/home/ubuntu/girder/girder/external/mongodb_proxy.py", line 64, in __call__
    val = self.method(*args, **kwargs)
  File "/home/ubuntu/env/local/lib/python2.7/site-packages/pymongo/collection.py", line 849, in replace_one
    collation=collation)
  File "/home/ubuntu/env/local/lib/python2.7/site-packages/pymongo/collection.py", line 766, in _update
    codec_options=self.__write_response_codec_options).copy()
  File "/home/ubuntu/env/local/lib/python2.7/site-packages/pymongo/pool.py", line 482, in command
    self._raise_connection_failure(error)
  File "/home/ubuntu/env/local/lib/python2.7/site-packages/pymongo/pool.py", line 610, in _raise_connection_failure
    raise error
InvalidStringData: strings in documents must be valid UTF-8: 'Pigmented Spitz\x82s Nevus'
```

msmolens:
Require metadata CSV files to be UTF-8 encoded. Validation of CSV files
that aren't UTF-8 encoded will indicate an error that provides details
of the first encoding problem encountered in the file.

Because MongoDB stores data in BSON format and BSON strings are UTF-8
encoded, strings stored to the database must contain valid UTF-8 data
[1]. Validating that the CSV file is UTF-8 encoded avoids errors when
saving strings to the database later in the workflow. For example:

    InvalidStringData: strings in documents must be valid UTF-8

The implementation replaces the csv module with a backport of Python 3's
csv module [2]. This ensures that csv.DictReader properly reads UTF-8
CSV files and returns Unicode strings. Additionally, using this backport
instead of managing UTF-8 conversions manually should make a future
transition to Python 3 easier.

Note that the implementation currently isn't compatible with Python 3
because of the use of the unicode() function.

[1] http://api.mongodb.com/python/current/tutorial.html#a-note-on-unicode-strings
[2] https://github.com/ryanhiebert/backports.csv

Fixes #473

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+backports.csv>=1.0.5
 geojson>=1.3.2
 python-dateutil>=2.6.0
 numpy>=1.10.2
--- a/plugin_tests/upload_test.py
+++ b/plugin_tests/upload_test.py
@@ -350,17 +350,17 @@ class UploadTestCase(IsicTestCase):
         self.assertEqual(
             resp.json['errors'], [
                 {'description':
-                 'on CSV row 5: values [\'solar lentigo\', False] for fields [\'diagnosis\', '
+                 'on CSV row 5: values [u\'solar lentigo\', False] for fields [\'diagnosis\', '
                  '\'melanocytic\'] are inconsistent'}
             ])
         self.assertEqual(
             resp.json['warnings'], [
                 {'description':
-                 'on CSV row 4: no images found that match \'filename\': \'test_1_small_3.jpg\''},
+                 'on CSV row 4: no images found that match u\'filename\': u\'test_1_small_3.jpg\''},
                 {'description':
-                 'on CSV row 6: no images found that match \'filename\': \'test_1_large_2.jpg\''},
+                 'on CSV row 6: no images found that match u\'filename\': u\'test_1_large_2.jpg\''},
                 {'description':
-                 'unrecognized field \'age_approx\' will be added to unstructured metadata'},
+                 'unrecognized field u\'age_approx\' will be added to unstructured metadata'},
                 {'description':
-                 'unrecognized field \'isic_source_name\' will be added to unstructured metadata'}
+                 'unrecognized field u\'isic_source_name\' will be added to unstructured metadata'}
             ])
--- a/server/models/dataset.py
+++ b/server/models/dataset.py
@@ -17,7 +17,6 @@
 #  limitations under the License.
 ###############################################################################
 
-import csv
 import datetime
 import itertools
 import os
@@ -28,6 +27,7 @@ from girder.models.model_base import GirderException, ValidationException
 from girder.models.notification import ProgressState
 from girder.utility import assetstore_utilities, mail_utils
 from girder.utility.progress import ProgressContext
+from backports import csv
 
 from .dataset_helpers import matchFilenameRegex
 from .dataset_helpers.image_metadata import addImageMetadata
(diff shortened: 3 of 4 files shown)


brianhelba:
@msmolens Can you rebase this onto master, so tests can pass?

It makes sense that without `backports.csv`, a UTF-8-encoded CSV would not yield `unicode` strings (explaining the original bug).

However, are you aware of any instances of CSVs with Unicode characters which were not encoded in UTF-8, or is the `except UnicodeDecodeError as e:` just there as a proactive guard?

msmolens:
> without backports.csv, a UTF-8-encoded CSV would not yield unicode strings (explaining the original bug).

Not exactly. The first reference link in the commit message explains PyMongo's behavior. In particular:

>  PyMongo must ensure that any strings it stores contain only valid UTF-8 data. Regular strings (<type ‘str’>) are validated and stored unaltered. Unicode strings (<type ‘unicode’>) are encoded UTF-8 first

The original bug occurred when the regular strings we input to PyMongo are validated and those strings aren't valid UTF-8.

> However, are you aware of any instances of CSVs with Unicode characters which were not encoded in UTF-8, or is the except UnicodeDecodeError as e: just there as a proactive guard?

Yes, the CSVs on the live site that demonstrated the bug are not valid UTF-8. While we could use heuristics to guess the encoding, it seems far better to require a single standard encoding. Some education or additional documentation might be helpful for dataset contributors. I tested re-saving the problem CSVs using Google Sheets, which saves CSV files in UTF-8, and the updated CSVs work, following this change.

The combination of `backports.csv` and the `except UnicodeDecodeError` effectively moves the UTF-8 validation earlier in the workflow, while processing of the CSV stream. Additionally, using `backports.csv` enables reading UTF-8 CSV files that start with a BOM. With the Python 2.7 csv.DictReader, the BOM is erroneously read as part of the first field name.

brianhelba:
Ah, that makes sense. I forgot that PyMongo was able to take a `str` object that already contained UTF-8 encoded characters, in addition to `str` objects with plain Latin-1 characters (the ordinary case).

So:
* with the old behavior, a UTF-8 CSV would just go directly into a `str`, and through to PyMongo and the DB
* with the new behavior, a UTF-8 CSV will be deserialized into `unicode` (validating it in the process), then be serialized back into a UTF-8 `str`  by PyMongo

Is this correct?

requirements.txt
```diff
@@ -1,3 +1,4 @@
+backports.csv==1.0.5
```
brianhelba: Actually, can we make this a more abstract version specifier?

msmolens:
Yes, the two bullet points above describe the behavior as I understand it. With the old behavior, a `str` that's not valid UTF-8 results in the `InvalidStringData: strings in documents must be valid UTF-8` error when writing to the database.

requirements.txt
msmolens: Changed to `>=` with the same version number.

PR #486, opened 2017-11-01
Upgrade Girder

brianhelba:


brianhelba:
@msmolens Merging this now, as we can't build without it, due https://github.com/girder/girder/pull/2451 .

msmolens:
@brianhelba Thanks for fixing all the problems so quickly. ⚡️

PR #490, opened 2017-11-03
Upgrade large_image to a mainline version

brianhelba:


brianhelba:
@msmolens This shouldn't make any changes, it just takes us off the fork.

PR #493, opened 2017-11-06
Create sections for gallery facet types

brianhelba:


web_external/ImagesGallery/Facets/ImagesFacetsPane.js
```diff
@@ -35,27 +37,47 @@ const ImagesFacetsPane = View.extend({
...
-            this.$el.append(facetView.el);
-            // Do not render until the view has been inserted into the main DOM
-            facetView.render();
+        this._createFacetView('meta.clinical.diagnosis')
+            .$el.insertBefore(this.$('.isic-images-facets-clinical'));
+        this.completeFacets.forEach((completeFacet) => {
+            const facetId = completeFacet.id;
+
+            let headerEl = null;
+            // TODO: Use String.startswith and a ES6 polyfill
```
msmolens: Does `string.startsWith` not work?

brianhelba: It's a new feature of ES6. We don't provide a polyfill at the moment, so there's no reason to break unnecessarily until we add a polyfill or explicitly drop support for browsers without native ES6 support.

msmolens: I see, even with Babel using this would require a polyfill.

brianhelba: Yeah.

PR #496, opened 2017-11-07
 Add browser detection and warning message

brianhelba:
image


brianhelba:
@msmolens I'm not totally happy with how this is injecting content inside the top of `#g-app-body-container`, but that's the best way that I was able to get it to work with the fixed-position header. If we ever refactor the header's positioning, we should revisit the placement of this alert too.

brianhelba:
Note, I just added Edge to the blacklist too, since apparently some of the downstream multirater viewer is having a problem with it.

Issue #481, opened 2017-10-30
Boolean facet selection and filtering is broken


brianhelba:
Boolean facets are not performing correctly on the client-side.

brianhelba:
Working on this now. The API is fine.

brianhelba:
This is resolved.

PR #497, opened 2017-11-09
Fix faceted search for Melanocytic attribute

msmolens:


msmolens:
Sorry, didn't see #481. Feel free to close this if necessary.

brianhelba:
@msmolens Thanks, this is the fix for #481. I didn't have time to do it yet, so I'm glad you did.

Can you also apply this to the "family history" and "personal history" fields? Those are also booleans.

PR #501, opened 2017-11-10
Validate nevus type field

### LINKED ISSUE #479: Add metadata validation for "nevus_type"
`nevus_type`

Should only be set when `diagnosis` is `nevus` or `nevus spilus`.

Values can be:
* `Blue`
* `Combined`
* `Nevus NOS`
* `Deep Penetrating`
* `Halo`
* `Persistent/Recurrent`
* `Pigmented Spindle Cell of Reed`
* `Plexiform Spindle Cell`
* `Special Site`
* `Spitz`
* `Unknown`

msmolens:
Resolves #479

### DIFF
--- a/server/models/dataset_helpers/image_metadata.py
+++ b/server/models/dataset_helpers/image_metadata.py
@@ -481,24 +481,20 @@ class NevusTypeFieldParser(FieldParser):
             value = value.lower()
             if value in ['', 'unknown']:
                 value = None
-            elif value == 'not applicable':
-                value = None
             else:
-                # TODO: finish this
+                if value == 'nevus nos':
+                    value = 'nevus NOS'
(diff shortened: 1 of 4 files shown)


server/models/dataset_helpers/image_metadata.py
```diff
@@ -481,39 +481,25 @@ def transform(cls, value):
...
                 value = None
-            elif value == 'not applicable':
-                value = None
             else:
-                # TODO: finish this
                 cls._assertEnumerated(value, {
                     'blue',
                     'combined',
-                    'nevus NOS',
+                    'nevus nos',
```
brianhelba: In keeping with `"lentigo NOS"`, I think this should be `"nevus NOS"` too.

PR #502, opened 2017-11-20
Move Dataset to an isolated model type

brianhelba:


web_external/Datasets/datasetPage.pug
```diff
@@ -32,9 +32,6 @@
       tr
         td.isic-listing-item-table-label Owner
         td= dataset.get('owner')
-      tr
-        td.isic-listing-item-table-label Signature
-        td= dataset.get('signature')
   if dataset.canAdmin()
```
msmolens: Do these permission checks need to be updated? It looks like being a Dataset Contributor is insufficient to access the metadata endpoints.

brianhelba: Fixed.

web_external/Datasets/datasetInfoWidget.pug
```diff
@@ -11,6 +11,13 @@ table.table.table-condensed.isic-dataset-info-table
     tr
       td.isic-dataset-info-table-label License
       td= dataset.get('license')
+    tr
+      td.isic-dataset-info-table-label Attribution
+      td= dataset.get('attribution')
```
msmolens: Attribution isn't available for the dataset summary, so displaying it should be inside a conditional.

web_external/Datasets/datasetInfoWidget.pug
```diff
@@ -11,6 +11,13 @@ table.table.table-condensed.isic-dataset-info-table
...
+    if dataset.canWrite()
+      tr
+        td.isic-dataset-info-table-label Owner
+        td= dataset.get('owner')
```
msmolens: Same here.

web_external/Datasets/selectDatasetView.js
```diff
@@ -21,7 +21,10 @@ const SelectDatasetView = View.extend({
     datasetChanged: function () {
         const datasetId = this.$('select').val();
         const dataset = this.collection.get(datasetId);
-        this.trigger('changed', dataset);
+        dataset.fetch()
+            .done(() => {
+                this.trigger('changed', dataset);
+            });
```
msmolens: Why is it necessary to fetch the dataset in this view?

msmolens: I see, so the details are available!

PR #508, opened 2017-12-22
Updates to the annotation API

brianhelba:


server/api/annotation.py
```diff
@@ -39,15 +39,20 @@ def __init__(self):
...
+        self.route('GET', (':annotationId', ':featureId', 'render'),
+                   self.getAnnotationFeatureRendered)
+        self.route('PUT', (':annotationId',), self.submitAnnotation)
 
     @describeRoute(
         Description('Return a list of annotations.')
         .param('studyId', 'The ID of the study to filter by.', paramType='query', required=True)
         .param('userId', 'The ID of the user to filter by.', paramType='query', required=False)
         .param('imageId', 'The ID of the image to filter by.', paramType='query', required=False)
+        .param('state', 'Filter annotations to those at a given state', paramType='query',
```
msmolens: Nit: The other parameter descriptions end in a period, so this one should too.

server/api/annotation.py
```diff
@@ -68,66 +73,102 @@ def find(self, params):
...
+    @access.cookie
+    @access.public
+    @loadmodel(map={'annotationId': 'annotation'}, model='annotation', plugin='isic_archive',
+               level=AccessType.READ)
+    def getAnnotationFeatureMask(self, annotation, featureId, params):
+        study = Study().load(annotation['meta']['studyId'], force=True, exc=True)
+        featureset = Study().getFeatureset(study)
+
+        if not any(featureId == feature['id'] for feature in featureset['localFeatures']):
+            raise ValidationException('Invalid featureId.', 'featureId')
```
msmolens: It may be of interest that, when `produces` is defined above, swagger-ui never indicates this error response, even though the request completes.

brianhelba: Reported upstream https://github.com/girder/girder/issues/2553

server/api/annotation.py
brianhelba: Fixing.

PR #509, opened 2018-01-04
Fix a bug when default values are accepted creating dataset

brianhelba:


web_external/Datasets/CreateDatasetView.js
```diff
@@ -23,19 +23,8 @@ const CreateDatasetView = View.extend({
...
-            let anonymous = this.$('#isic-create-dataset-attribution-anonymous').prop('checked');
-            const attribution = this.$('#isic-create-dataset-attribution-name').val();
-            if (_.contains(['anonymous', 'anon'], attribution.toLowerCase())) {
-                // This could lead to a non-valid UI state, but it will be caught by server-side
-                // validation and it's still easy for the user to resolve
-                anonymous = true;
-            }
-            this.dataset.set('attribution', anonymous ? 'Anonymous' : attribution);
-        },
+        'change #isic-create-dataset-license': this._setLicense,
```
msmolens: This line, and below, should use a string instead, like `'_setLicense'`.

PR #511, opened 2018-01-10
Single image upload API

msmolens:
Remaining tasks could include:
- [ ] Send email confirmations of upload (?)


msmolens:
@brianhelba PTAL at the latest changes. I replaced the endpoint that worked like Girder's [POST /file/chunk](https://github.com/girder/girder/blob/e3b2f195c41c34f277e163153b20339448d93e19/girder/api/v1/file.py#L211-L213) endpoint with a two-step process. Now the image parameters (filename, signature) are sent separately from the image data.

msmolens:
@brianhelba PTAL

PR #517, opened 2018-02-01
Add an 'imageIds' query string option for images endpoints

brianhelba:
* Remove the 'datasetId' query string option from images endpoints
  * This same functionality can be accomplished more generally via a "filter" option.
* Add an 'imageIds' query string option for images endpoints

### DIFF
--- a/server/api/image.py
+++ b/server/api/image.py
@@ -73,6 +73,34 @@ class ImageResource(IsicResource):
         except (TypeError, ValueError) as e:
             raise ValidationException('Could not parse filter:' % str(e), 'filter')
 
+    def _parseImageIds(self, imageIdsParam):
+        if isinstance(imageIdsParam, six.string_types):
+            try:
+                imageIdsParam = json.loads(imageIdsParam)
+            except ValueError as e:
+                raise ValidationException('Could not decode JSON: %s' % str(e), 'imageIds')
+        if not isinstance(imageIdsParam, list):
+            raise ValidationException('"imageIds" must be a JSON-encoded array.', 'imageIds')
+        if len(imageIdsParam) > 100:
+            raise ValidationException('"imageIds" may not contain more than 100 IDs.', 'imageIds')
+        imageIds = []
+        for imageId in imageIdsParam:
+            if not isinstance(imageId, six.string_types):
+                raise ValidationException('Each of "imageIds" must be a string.', 'imageIds')
+            try:
+                imageId = ObjectId(imageId)
+            except InvalidId:
+                raise ValidationException('Invalid ObjectId.', 'imageIds')
+            imageIds.append(imageId)
+        # TODO: We could validate that the user has access to all of the requested IDs, and
+        # raise a 403 error if now. Currently, we will silently exclude images that the user doesn't
+        # have permission to READ.
+
+        query = {
+            '_id': {'$in': imageIds}
+        }
+        return query
+
     @describeRoute(
         Description('Return a list of lesion images.')
         .pagingParams(defaultSort='name')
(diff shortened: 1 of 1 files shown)


server/api/image.py
```diff
@@ -165,6 +201,9 @@ def metadataGenerator():
                enum=['all', 'images', 'metadata'], default='all')
         .param('filter', 'Filter the images by a PegJS-specified grammar.',
                required=False)
+        .param('imageIds',
+               'A JSON array of up to 100 image IDs, which will be exclusively included.',
+               required=False)
```
msmolens: This is a great candidate for [jsonParam](http://girder.readthedocs.io/en/latest/api-docs.html?highlight=jsonParam#girder.api.describe.Description.jsonParam)'s schema validation. Most of the validation in `_parseImageIds()` could be removed, with the example of the ObjectId check.

Here's an example from Covalic:
https://github.com/girder/covalic/blob/1c4805be1f5c72346220bb6d0dc87a9be14e7354/server/rest/submission.py#L359-L387

server/api/image.py
```diff
@@ -170,9 +199,10 @@ def metadataGenerator():
         Description('Download multiple images as a ZIP file.')
         .param('include', 'Which types of data to include.', required=False,
                enum=['all', 'images', 'metadata'], default='all')
-        .param('datasetId', 'The ID of the dataset to download.', required=False)
-        .param('filter', 'Filter the images by a PegJS-specified grammar (causing "datasetId" to '
-                         'be ignored).',
+        .param('filter', 'Filter the images by a PegJS-specified grammar.',
+               required=False)
+        .param('imageIds',
+               'A JSON array of up to 100 image IDs, which will be exclusively included.',
```
msmolens: One solution I've seen discussed to deal with a large number of query parameters is to use a POST request instead. Has this been considered?

server/api/image.py
brianhelba: Agreed, but we'd need to use `autoDescribeRoute` in order for it to have any effect. IMO, `autoDescribeRoute` is much too complicated to understand via its documentation alone (and I don't trust the behavior enough to copy-paste from upstream usages), so I'd rather move to that as part of a larger pass.

server/api/image.py
brianhelba: By my rough math, we could probably handle 300-500 imageIDs without too much of an impact (perhaps we'd have to just change [Nginx's `large_client_header_buffers` option](http://nginx.org/en/docs/http/ngx_http_core_module.html#large_client_header_buffers). Beyond that, we'd also want to consider how such a large query was being sent to MongoDB, even if we could get to the server via POST or otherwise. Additionally, if the images are being hand selected via a web client, the client itself would need to track and display the selection state of many individual images, which does not scale well either.

So, I think having some limit on the number of images is reasonable in general. I think we could safely bump up the 100 number in the future, but I don't foresee this ever allowing truly large numbers of selections.

Accordingly, I think we can keep the purity of using a `GET` here for now. I'd be fine to revisit this if the scaling requirements change in the future.

brianhelba:
See my responses. PTAL.

server/api/image.py
msmolens: It makes sense to do it as part of a larger pass. I already used it once in this project to take advantage of upstream JSON validation.
https://github.com/ImageMarkup/isic-archive/blob/5dc99980bf2b1be7f2cadb99b3a4a8b93475a3ee/server/api/image.py#L378

server/api/image.py
msmolens: Thanks for explaining the rationale. This all makes sense to me.

PR #518, opened 2018-02-01
Allow some metadata values to be "not applicable"

brianhelba:


server/models/dataset_helpers/image_metadata.py
```diff
@@ -705,7 +705,7 @@ def transform(cls, value):
         if value is not None:
             value = value.strip()
             value = value.lower()
-            if value in ['', 'unknown']:
+            if value in ['', 'unknown', 'not applicable']:
```
msmolens: This comment could be updated to include "not applicable" in the list of values, when, well, applicable.
https://github.com/ImageMarkup/isic-archive/blob/508951c0470870f1d89b7f135af0f14ce33fc633/server/models/dataset_helpers/image_metadata.py#L124-L125

brianhelba: Fixed.

PR #519, opened 2018-02-01
Support study participation requests

msmolens:


server/models/study.py
```diff
@@ -176,6 +186,54 @@ def getImages(self, study):
             'meta.studyId': study['_id']}).distinct('meta.imageId')
         return Image().find({'_id': {'$in': imageIds}})
 
+    def addWaitingUser(self, study, user):
+        """
+        Add user to list of users waiting to participate in the study.
+        """
+        waitingUsers = study['meta'].get('waitingUsers', [])
```
brianhelba: We can assume this always exists (here, and in other methods). I'll run a migration to add the field.

server/models/study.py
```diff
@@ -176,6 +186,54 @@ def getImages(self, study):
...
+        waitingUsers.append(user['_id'])
+
+        # "setMetadata" will always save
+        self.setMetadata(
```
brianhelba: I'd like to move away from `setMetadata`. Can you just do a `.save` or even a `.update` (with `$addToSet`)?

server/models/study.py
```diff
@@ -176,6 +186,54 @@ def getImages(self, study):
...
+    def removeWaitingUser(self, study, user):
+        """
+        Remove user from list of users waiting to participate in the study.
+        """
+        try:
+            waitingUsers = study['meta'].get('waitingUsers', [])
+            waitingUsers.remove(user['_id'])
+
+            # "setMetadata" will always save
+            self.setMetadata(
```
brianhelba: I'd like to move away from `setMetadata`. Can you just do a `.save` or even a `.update` (with `$pull`)?

server/models/study.py
```diff
@@ -176,6 +186,54 @@ def getImages(self, study):
...
+                folder=study,
+                metadata={
+                    'waitingUsers': waitingUsers
+                }
+            )
+        except ValueError:
```
brianhelba: Can you scope the associated `try` block for this `ValueError` to a more specific statement?

web_external/models/StudyModel.js
```diff
@@ -33,21 +33,28 @@ const StudyModel = Model.extend({
...
     },
 
+    waitingUsers: function () {
+        let userModels;
+        if (this.has('waitingUsers')) {
+            userModels = this.get('waitingUsers').map((user) => {
+                return new UserModel(user);
+            });
+        } else {
+            userModels = [];
```
brianhelba: If someone is calling this method and they're not an admin, I'd rather it produce an error / failure, instead of silently returning an empty list.

web_external/Studies/studyPage.pug
```diff
@@ -41,12 +41,25 @@
...
                 td
                   if study.canAdmin()
-                    a.isic-study-remove-user-button(title='Delete annotator')
+                    a.isic-study-remove-user-button.isic-study-delete-button(title='Delete annotator')
                       i.icon-cancel
                 td.isic-study-user-completion-count= numCompleted
                 td.isic-study-user-completion-percent= percentCompleted
         else
           | (no users)
+      if study.canAdmin() && study.get('waitingUsers').length
```
brianhelba: Can we always show this field to admins, even if there are currently no waiting users (to train admins to keep in mind that this functionality exists)?

Maybe show a fallback placeholder if the list is empty.

server/models/study.py
```diff
@@ -280,6 +341,9 @@ def filter(self, study, user=None, additionalKeys=None):
             }
         }
 
+        if User().canAdminStudy(user):
+            output['waitingUsers'] = getSortedUserList(Study().getWaitingUsers(study))
```
brianhelba: To eventually accommodate a feature where admins issued soft "invites" for users to participate in a study as annotators, I think this should be given a more precise name than `waitingUsers`. I can't think of a good one right now though.

msmolens: How about `requests`? This avoids using a long name and matches a similar field in Girder's group model.

brianhelba: Honestly, that seems a little ambiguous still, since an admin could also issue "requests" for users to join. I'm fine with longer / compound names, I don't see big advantages to being overly terse.

There are only two hard problems in computer science...

PR #525, opened 2018-02-08
Metadata: Autocorrect melanocytic field inconsistencies

### LINKED ISSUE #524: Auto-correct "melanocytic" metadata inconsistancies
* `melanocytic` validation failures (w.r.t `diagnosis`) should emit a warning only
* `melanocytic` validation failures should be auto-corrected (based on `diagnosis`)
  * Since these are so trivial to auto-correct (given a diagnosis) and are such a common human error, we will auto-correct inconsistencies, favoring the value derived from the diagnosis field

msmolens:
Validation failures for the melanocytic field with respect to the diagnosis field are now autocorrected and a warning is generated.

Note that the metadata logic still prevents existing non-null melanocytic values from being overwritten.

Fixes #524

### DIFF
--- a/server/models/dataset_helpers/image_metadata.py
+++ b/server/models/dataset_helpers/image_metadata.py
@@ -720,13 +720,18 @@ class MelUlcerFieldParser(FieldParser):
 def _populateMetadata(acquisition, clinical):
     """
     Populate empty metadata fields that can be determined based on other fields.
+    In some cases, populates inconsistent fields and emits a warning.
+    Returns a list of warnings.
     """
+    warnings = []
+
     dermoscopicType = acquisition.get('dermoscopic_type')
     imageType = acquisition.get('image_type')
 
     diagnosis = clinical.get('diagnosis')
     benignMalignant = clinical.get('benign_malignant')
     diagnosisConfirmType = clinical.get('diagnosis_confirm_type')
+    melanocytic = clinical.get('melanocytic')
 
     if diagnosis == 'melanoma':
         if benignMalignant is None:
@@ -737,14 +742,28 @@ def _populateMetadata(acquisition, clinical):
 
     # Set melanocytic field based on diagnosis
     if diagnosis in MelanocyticFieldParser.melanocyticDiagnoses:
-        if clinical.get('melanocytic') is None:
+        if melanocytic is None:
             clinical['melanocytic'] = True
+        elif not melanocytic:
+            clinical['melanocytic'] = True
+            warnings.append('corrected inconsistent value for field %r based on field %r '
+                            '(new value: %r, %r: %r)' %
+                            (MelanocyticFieldParser.name, DiagnosisFieldParser.name,
+                             True, DiagnosisFieldParser.name, diagnosis))
     elif diagnosis is not None and diagnosis != 'other':
-        if clinical.get('melanocytic') is None:
+        if melanocytic is None:
             clinical['melanocytic'] = False
+        elif melanocytic:
+            clinical['melanocytic'] = False
+            warnings.append('corrected inconsistent value for field %r based on field %r '
+                            '(new value: %r, %r: %r)' %
+                            (MelanocyticFieldParser.name, DiagnosisFieldParser.name,
+                             False, DiagnosisFieldParser.name, diagnosis))
 
     if dermoscopicType is not None and imageType is None:
         acquisition['image_type'] = 'dermoscopic'
+
+    return warnings
 
 
 def _checkMetadataErrors(acquisition, clinical):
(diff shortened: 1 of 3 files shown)


server/models/dataset_helpers/image_metadata.py
```diff
@@ -737,15 +741,31 @@ def _populateMetadata(acquisition, clinical):
 
     # Set melanocytic field based on diagnosis
     if diagnosis in MelanocyticFieldParser.melanocyticDiagnoses:
-        if clinical.get('melanocytic') is None:
+        melanocytic = clinical.get('melanocytic')
```
brianhelba: Is there a reason not to put this statement outside the conditional, so it can be shared by both branches?

server/models/dataset_helpers/image_metadata.py
```diff
@@ -737,15 +741,31 @@ def _populateMetadata(acquisition, clinical):
...
+        if melanocytic is None:
             clinical['melanocytic'] = False
+        elif melanocytic:
+            clinical['melanocytic'] = False
+            warnings.append('corrected inconsistent value for field %r based on field %r '
+                            '(new value: %r, %r: %r)' %
+                            (MelanocyticFieldParser.name, DiagnosisFieldParser.name,
+                             True, DiagnosisFieldParser.name, diagnosis))
```
brianhelba: The new value logged should probably be `False`.

server/models/dataset_helpers/image_metadata.py
```diff
@@ -737,15 +741,31 @@ def _populateMetadata(acquisition, clinical):
...
     if diagnosis in MelanocyticFieldParser.melanocyticDiagnoses:
-        if clinical.get('melanocytic') is None:
+        melanocytic = clinical.get('melanocytic')
+        if melanocytic is None:
+            clinical['melanocytic'] = True
+        elif not melanocytic:
             clinical['melanocytic'] = True
+            warnings.append('corrected inconsistent value for field %r based on field %r '
+                            '(new value: %r, %r: %r)' %
+                            (MelanocyticFieldParser.name, DiagnosisFieldParser.name,
```
brianhelba: As these template strings get more complicated, it might be helpful to switch to passing a `dict` as the injected values. Then you can refer to values as `%(keyName)s` instead of just `%s`.

server/models/dataset_helpers/image_metadata.py
```diff
@@ -720,7 +720,11 @@ def load(cls, value, acquisition, clinical, private):
 def _populateMetadata(acquisition, clinical):
     """
     Populate empty metadata fields that can be determined based on other fields.
+    In some cases, populates inconsistent fields and emits a warning.
+    Returns a list of warnings.
     """
+    warnings = []
```
brianhelba: In the future (I think it's out of scope of this PR), we could consider using the [built-in `warnings` library](https://docs.python.org/2.7/library/warnings.html) for this.

server/models/dataset_helpers/image_metadata.py
msmolens: Fixed.

server/models/dataset_helpers/image_metadata.py
msmolens: Fixed.

server/models/dataset_helpers/image_metadata.py
msmolens: Since that requires specifying dictionary keys, this seems less verbose. But I'll consider that in the future.

server/models/dataset_helpers/image_metadata.py
msmolens: I'm not clear how that library is appropriate for an application-level list of strings that happen to be warnings about the data.

server/models/dataset_helpers/image_metadata.py
brianhelba: Yeah, it's a trade-off, for slightly more typing, in exchange for less reliance of magical positioning of variables within the template string.

server/models/dataset_helpers/image_metadata.py
brianhelba: I think we'd use a [`warnings.catch_warnings`](https://docs.python.org/2.7/library/warnings.html#warnings.catch_warnings) context manager around our application logic. I'm not certain this is ultimately a good idea, just that we might want to consider it (so we don't have to keep passing and returning `warnings` variables).

msmolens: Ah, I see. I agree that catch_warnings, or a similar construct, would improve the maintainability of this code. Here's an example of what that might look like: ://docs.python.org/3/library/warnings.html#testing-warnings

PR #526, opened 2018-02-08
Document additional operations for user lifecycle management in Swagger

brianhelba:


server/__init__.py
```diff
@@ -112,10 +112,30 @@ def replaceHost(url):
...
+        ('GET', '/oauth/provider'),  # initiate an OAuth2 workflow for login / registration
+        ('PUT', '/user/{id}'),  # change personal info
+        ('PUT', '/user/password'),  # change password
+        ('PUT', '/user/password/temporary'),  # initiate email reset of forgotten password
+        ('GET', '/user/password/temporary/{id}'),  # complete email reset of forgotten password
+        ('POST', '/user/verification'),  # resend an email verification message
+    }:
+        # The [0] element of the routePath split is '', since it starts with a '/'
+        routeResource = routePath.split('/')[1]
+        routeOperation = routes[routeResource][routePath][routeMethod]
```
msmolens: This line errors if the oauth plugin isn't enabled. Can you make it more robust?

brianhelba: Yeah, I'll fix.

PR #529, opened 2018-02-19
Solar lentigo is no longer considered melanocytic

msmolens:
This updates metadata validation. The database metadata should be updated separately.


brianhelba:
I'll run a migration before making this live.

PR #531, opened 2018-02-21
Add usage notes to Image apply metadata endpoint

msmolens:


server/api/image.py
```diff
@@ -407,6 +407,39 @@ def doSegmentation(self, image, params):
...
+               'is not saved, regardless of the value of the `save` parameter, and the `errors` '
+               'list contains the validation errors. Validation can fail for several reasons, '
+               'including but not limited to:\n'
+               '- The value of a field is of the wrong type\n'
+               '- The values of two or more fields are inconsistent with one another\n'
+               '- A value of a field already exists and cannot be safely overwritten\n'
+               '\n'
+               'When there are no errors, the `warnings` list contains warnings that don\'t '
+               'prevent saving the metadata, but should be reviewed.')
         .modelParam('id', model=Image, destName='image', level=AccessType.READ)
```
brianhelba: Can you put a `description` in this `.modelParam` call, to make it read `"The ID of the image."`. By default, it says `"The ID of the document."`, which is less clear.

msmolens: The upstream behavior changed but it still documented as automatically generating an appropriate description: https://github.com/girder/girder/blob/ffc49c0c5f92b4b4372512b015d24e83c4f4ef47/girder/api/describe.py#L295-L296
This is evident on sites like data.kitware.com: https://data.kitware.com/api/v1#!/file/file_getFile

PR #536, opened 2018-03-01
Upgrade GeoJS

brianhelba:
This should fix mouse interaction bugs on Firefox.


brianhelba:
This includes an extra commit to work around the issue https://github.com/OpenGeoscience/geojs/issues/786

If https://github.com/OpenGeoscience/geojs/pull/787 is merged and released soon, I'll update this PR to use a newer release of GeoJS, without this workaround.

PR #537, opened 2018-03-02
Separate annotation responses and markups

brianhelba:
https://github.com/ImageMarkup/isic-archive/wiki/Annotation provides the model for this change.

This will require another database migration, which is included.

This updates the `PUT /annotation/:annotationId` API in an incompatible way, as that's not used publicly. The annotation client is updated accordingly.

This renames other annotation-related APIs, deprecating the old names:
* get the markup as a PNG mask:
  * old: `GET /annotation/:annotationId/:featureId/mask`
  * new: `GET /annotation/:annotationId/markup/:featureId`
  * note, this is now the preferred way to get markup, as it will be generalizable to non-superpixel markup and does not require any special superpixel knowledge or masks
* get the markup as a rendered JPEG overlay on the image:
  * old: `GET /annotation/:annotationId/:featureId/render`
  * new: `GET /annotation/:annotationId/markup/:featureId/rendered`
* get the markup raw superpixels as JSON:
  * old: `GET /annotation/:annotationId/:featureId`
  * new: `GET /annotation/:annotationId/markup/:featureId/superpixels`

This also adds new fields to the `GET /annotation/:annotationId` API: `responses` and `markup`; however, the `annotations` field is still returned (though it's deprecated) and contains the concatenation of both `responses` and `markup`.

### DIFF
--- a/server/models/annotation.py
+++ b/server/models/annotation.py
@@ -49,7 +49,8 @@ class Annotation(AccessControlMixin, Model):
             'startTime': None,
             'stopTime': None,
             'status': None,
-            'annotations': None
+            'responses': {},
+            'markups': {},
         })
 
         return annotation
@@ -63,7 +64,7 @@ class Annotation(AccessControlMixin, Model):
         if self.getState(annotation) != Study().State.COMPLETE:
             raise GirderException('Annotation is incomplete.')
 
-        featureValues = annotation['annotations'].get(featureId, [])
+        featureValues = annotation['markups'].get(featureId, [])
         if not isinstance(featureValues, list):
             raise GirderException(
                 'Feature %s is not a superpixel annotation.' % featureId)
@@ -98,7 +99,7 @@ class Annotation(AccessControlMixin, Model):
 
         return possibleMask, definiteMask
 
-    def renderMask(self, annotation, featureId):
+    def maskMarkup(self, annotation, featureId):
         possibleMask, definiteMask = self._getImageMasks(annotation, featureId)
 
         renderedMask = numpy.zeros(possibleMask.shape, dtype=numpy.uint)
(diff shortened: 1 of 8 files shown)


brianhelba:
@msmolens PTAL. I need to fully test this myself, but this is feature-complete.

After this, I think the next steps are:
* Move markup values to their own `File` models
  * These will no longer have to render on-demand, speeding up the server
  * This will remove the huge superpixel value arrays from annotation documents, allowing for real-time aggregation of annotation documents
* Migrate to the new master feature list
  * It'd be nice to have a more graceful API migration / deprecation pathway for this, but I'm not sure what it is yet
  * Our ability to remove featuresets will depend on the annotation tool; there may be a temporary way to accept new annotations from the old featuresets on existing studies (to avoid disruption to users)
* Extend the histogram aggregation pipeline to generate a markup facet
  * I have a working prototype of this, but it's currently too slow and meaningless without a master feature list
* Add support for image filtering by markup

server/models/annotation.py
```diff
@@ -49,7 +49,8 @@ def createAnnotation(self, study, image, user):
             'startTime': None,
             'stopTime': None,
             'status': None,
-            'annotations': None
+            'responses': {},
+            'markups': {},
```
msmolens: https://github.com/ImageMarkup/isic-archive/wiki/Annotation describes these as lists, but it looks like they're mappings. Should the wiki be clarified, including the example JSON"

brianhelba: They should be mappings, this code is incorrect. However, what's being set here is an incomplete annotation. Once an annotation is submitted, the whole field is overwritten, so this code doesn't really matter. However, I was thinking of initializing these fields as empty containers instead of `None` just to make aggregation development more convenient.

msmolens: Initializing them as empty containers makes sense. I meant we should clarify in the documentation whether they are of `list` or `dict` type. Currently the example on the wiki is invalid JSON that mixes list and object formats, e.g.:
```
"responses": [
    "Classify this lesion as benign or malignant": "Benign",
    "Enter any other clinical observations": "atypical nevus"
],
```

brianhelba: Got it, that's a typo.

PR #539, opened 2018-03-06
Return questions and features inline with study, in new format

brianhelba:


brianhelba:
@msmolens PTAL. I'm open to iterating on this new format.

Also, this completely elides the old "id" field of questions / features, which means that users won't have enough info to submit annotations if they only use these new fields. Perhaps as a transition mechanism, we could update the annotation submission API to accept responses / markup by name too (assuming the old feature names are all unique within a featureset). What do you think?

msmolens:
@brianhelba I'd prefer to maintain "id" fields for questions/features and responses. If user-readable text isn't used as a key it's easier to modify later.

brianhelba:
I've tested the annotation submission API changes, and they work.

brianhelba:
@msmolens Note the final commit here, which updates the `features` field in a Study to return each feature as an object, instead of a string. This is in anticipation of extra options (like `required`) being added to some features in the context of a study. I'll update the wiki accordingly.

PR #544, opened 2018-03-20
Fix a bug where batch creators were not stored properly

brianhelba:


brianhelba:
@msmolens It looks like the builds are broken until we pull down https://github.com/girder/girder/pull/2663 . I'll make a PR to upgrade Girder.

PR #545, opened 2018-03-21
Implement annotation tool in Vue

msmolens:
This replicates the existing annotation tool functionality using a modern framework. This improves maintainability and simplifies adding new features or a potential redesign.

Notes:
- eslint-plugin-vue currently doesn't support linting pug templates, only HTML.
- Vue code is often written using modern JavaScript syntax, in particular the object spread operator and async/await. Supporting those here would require additional Babel plugins and webpack configuration.
- This PR currently doesn't remove the legacy annotation tool code.


web_external/vue/views/AnnotationTool.vue
```diff
@@ -0,0 +1,280 @@
...
+      )
+</template>
+
+<script>
+import { createNamespacedHelpers } from 'vuex';
+
+import { getApiRoot } from 'girder/rest';
+import { getCurrentUser } from 'girder/auth';
+
+import AnnotationToolGlobalFeatures from './AnnotationToolGlobalFeatures.vue';
```
brianhelba: Can we rename these classes and files to refer to `Questions` instead of `GlobalFeatures`?

web_external/vue/views/AnnotationTool.vue
```diff
@@ -0,0 +1,280 @@
...
+import AnnotationToolLocalFeatures from './AnnotationToolLocalFeatures.vue';
```
brianhelba: Can we rename these classes and files to refer to `Features` instead of `LocalFeatures`?

web_external/vue/views/AnnotationTool.vue
```diff
@@ -0,0 +1,280 @@
+<template lang="pug">
+.annotation-tool-container.container-fluid
+  .isic-page.row
+    .isic-pane.col-sm-3
+      .header
+        .info-group
```
brianhelba: This group is not necessary.

web_external/vue/views/AnnotationTool.vue
```diff
@@ -0,0 +1,280 @@
...
+          span.title ISIC Annotation Tool
+          a.pull-right(href='/') Return home.
+        .info-group
+          div(v-if='image')
+            span #[b Image:]&nbsp;{{ image.name }}
+        .info-group
```
brianhelba: This group is not necessary.

web_external/vue/views/AnnotationTool.vue
```diff
@@ -0,0 +1,280 @@
...
+    ]), mapActions([
+        'getAnnotation',
+        'submitAnnotation'
+    ]))
+};
+</script>
+
+<style lang="stylus" scoped>
+.annotation-tool-container
+  background #000
```
brianhelba: We don't need to override with a black background for any of these components. We can inherit the color scheme of the main site, for the most part.

web_external/vue/views/AnnotationTool.vue
```diff
@@ -0,0 +1,280 @@
...
+                a(@click='flag("other")') Other reason
+            span.flag-status
+              span #[b Flag Status:]&nbsp;{{ flagStatus }}
+      .annotation(v-if='questions.length > 0')
+        h3.annotation-header Global Features
+        annotation-tool-global-features
+      .annotation(v-if='questions.length > 0')
+        h3.annotation-header Local Features
+        .annotation-section
+          annotation-tool-local-features(
```
brianhelba: Should this use the camelcased component name?

web_external/vue/views/AnnotationTool.vue
```diff
@@ -0,0 +1,280 @@
...
+              li
+                a(@click='setFlagStatus("inappropriate")') Clinically inappropriate
+              li.divider
+              li
+                a(@click='flag("other")') Other reason
+            span.flag-status
+              span #[b Flag Status:]&nbsp;{{ flagStatus }}
+      .annotation(v-if='questions.length > 0')
+        h3.annotation-header Global Features
+        annotation-tool-global-features
```
brianhelba: Should this use the camelcased component name?

web_external/vue/views/AnnotationTool.vue
msmolens: I preserved the old look and feel to focus on the functionality instead of a redesign. I think inheriting the white site background forces use to reconsider the style of everything to be usable.

web_external/vue/views/AnnotationTool.vue
msmolens: As I understand it, no, it has to be kebab-case.
https://vuejs.org/v2/guide/components.html#Component-Naming-Conventions

web_external/vue/views/AnnotationTool.vue
msmolens: Removing these grounds  changes the style--specifically padding--compared to the original implementation.

web_external/vue/views/AnnotationTool.vue
msmolens: I'll remove this group.

web_external/vue/components/AnnotationTool/AnnotationToolFeatures.vue
```diff
@@ -0,0 +1,172 @@
...
+    label Confidence&nbsp;
+    .btn-group(data-toggle='buttons')
+      label.annotation-radio-btn.btn.btn-default(
+        :class='{active: markupState === 1.0}'
+      )
+        input(
+          type='radio',
+          name='certainty',
+          :value='1.0',
+          :checked='markupState === 1.0',
```
brianhelba: What do you think about using a [Two-way Computed Property](https://vuex.vuejs.org/en/forms.html) here instead (or more generally)? It would have avoided the bug you fixed.

msmolens: These radios used v-model in my initial implementation--before Vuex--but I changed that at some point. Now that this uses Vuex, I updated this to use a two-way computed property. It's much cleaner. Thanks!

brianhelba: @msmolens You can remove the `onCertaintyChanged` now too, right?

brianhelba:
LGTM!

PR #547, opened 2018-03-21
Add new metadata values

brianhelba:


server/models/dataset_helpers/image_metadata.py
```diff
@@ -361,7 +361,8 @@ def transform(cls, value):
                 cls._assertEnumerated(value, {
                     'histopathology',
                     'serial imaging showing no change',
-                    'single image expert consensus'})
+                    'single image expert consensus',
+                    'confocal with consensus dermoscopy'})
```
msmolens: Should the new value be added to the single image upload tool as well?

brianhelba: Good point.

brianhelba: Fixed.

brianhelba:
@msmolens Just updated this again. Ready for final review.

PR #548, opened 2018-03-29
Improve dataset license support

brianhelba:


brianhelba:
It's intentional that we don't expose GUI options for `CC-BY-NC-SA` licenses just yet, while we continue to refine the policy for when this license type should be accepted.

PR #555, opened 2018-04-25
Segmentation Review

msmolens:
Add a view for segmentation experts to approve/reject segmentations.

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -21,6 +21,7 @@
             "url-loader": "~0.6.2",
             "vue": "~2.5.0",
             "vue-loader": "~14.1.0",
+            "vue-simple-spinner": "~1.2.0",
             "vue-template-compiler": "~2.5.0",
             "vue2-dropzone": "~3.0.0",
             "vuex": "~3.0.0"
--- a/server/api/task.py
+++ b/server/api/task.py
@@ -311,7 +311,7 @@ class TaskResource(IsicResource):
 
         results = list(Image().collection.aggregate(pipeline))
         if not results:
-            raise RestException('No segmentations are needed for this dataset.')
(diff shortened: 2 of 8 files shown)


msmolens:
@brianhelba The view is currently accessible by navigating to `/#tasks/segmentationreview/<datasetId>`.
I propose adding a “Segmentation Review” group to Task Dashboard that lists each dataset with segmentations that require review.

The initial implementation reuses `/task/me/segmentation/next` to get images with segmentations to review, but I think we need to add a new endpoint with the following behavior:

- Return values only for segmentation experts
- Use an aggregation pipeline to select images that
  - Have at least one segmentation
  - Have no segmentations approved by a segmentation expert

The endpoints could be named like:
-  `GET /task/me/segmentationreview`
-  `GET /task/me/segmentationreview/next`

Does these changes sound reasonable?

brianhelba:
After reviewing this in-person, let's:
* Move the header bar into the left column, to allow the right column access to the full vertical space
* Move the keyboard help tooltip down to its own row, to ensure a sane layout on smaller displays
* Swap the keyboard shortcuts to make the leftward button (on a QWERTY keyboard) reject and the rightward button accept
* Add the spacebar as another keyboard shortcut for accept
* File a bug to document the issue with button-styled radio buttons not working with Vue + Bootstrap
* When confirm gets activated via keybinding, ensure the focus is cleared from previously-clicked accept / reject buttons
* Disable the confirm button element when accept / reject have not yet been selected

msmolens:
@brianhelba The interface changes from your review are complete.

brianhelba:
Let's merge this now, just to get the code available, then handle the task routing subsequently.

PR #556, opened 2018-04-27
Store markups in their own mask files

brianhelba:


scripts/migrateAnnotations.py
```diff
@@ -0,0 +1,17 @@
...
+for study in Study().find():
+    print study['name']
+
+    for annotation in Annotation().find({'studyId': study['_id']}):
+        print ' ', annotation['_id']
+
+        oldSuperpixelsMarkups = annotation['markups']
+
+        annotation['markups'] = {}
+        for featureId, superpixelValues in oldSuperpixelsMarkups:
```
msmolens: This gives me `ValueError: too many values to unpack` because `annotation['markups']` is a `dict`.

brianhelba: Should have a `.items()`. Will fix.

server/api/annotation.py
```diff
@@ -152,31 +171,35 @@ def getAnnotationMarkupSuperpixels(self, annotation, featureId, params):
...
-            'inline; filename="%s"' % contentName)
-        setResponseHeader('Content-Length', len(renderEncodedData))
+        if markupFile:
+            return File().download(markupFile, headers=True, contentDisposition='inline')
+        else:
+            image = Image().load(annotation['imageId'], force=True, exc=True)
+            markupMask = numpy.zeros(
+                (
+                    image['meta']['acquisition']['pixelsX'],
+                    image['meta']['acquisition']['pixelsY']
```
msmolens: I think the image height should be the first dimension here.

server/api/annotation.py
```diff
@@ -255,6 +274,40 @@ def submitAnnotation(self, annotation, params):
...
-        Annotation().save(annotation)
+        # Before anything is saved, validate all the other (non-'markups') fields
+        Annotation().validate(annotation)
+
+        # Validate superpixel markups
+        study = Study().load(annotation['studyId'], force=True, exc=True)
+        image = Image().load(annotation['imageId'], force=True, exc=True)
+        superpixelsData = Image().superpixelsData(image)
+        maxSuperpixel = int(superpixelsData.max())
+        jsonschema.validate(
```
msmolens: For consistency, since this raises `jsonschema.exceptions.ValidationError`, it would be good to catch that and raise `ValidationException`.

server/models/annotation.py
```diff
@@ -174,53 +223,75 @@ def validate(self, doc):  # noqa - C901
...
-                        'Annotation has invalid feature "%s".' % featureId)
-                if not (
-                    isinstance(markup, list) and
-                    len(markup) == maxSuperpixel + 1 and
-                    all(superpixelValue in [0.0, 0.5, 1.0]
-                        for superpixelValue in markup)
-                ):
-                    raise ValidationException(
-                        'Annotation feature "%s" has invalid markup "%s".' % (featureId, markup))
+            jsonschema.validate(
```
msmolens: This could also raise `ValidationException` on error.

brianhelba:
@msmolens Fixed.

PR #560, opened 2018-05-09
Fix a few issues with saving markup

brianhelba:


brianhelba:
@msmolens Merging now, but feel free to let me know any feedback.

PR #562, opened 2018-05-10
Separate EXIF metadata

### LINKED ISSUE #463: EXIF-related fields in CSV metadata should be segregated
Any fields beginning with `exif-` in CSV metadata should be stored in a separate place in the database (instead of the default under `meta.unstructured`). Two factors need to be considered in finding and naming an appropriate location:
* This is still unvalidated metadata, so it likely doesn't belong under `meta.acquisition` (although it is acquisition-related metadata).
* We may eventually dump EXIF tags directly from the the original image binary data. That will also be (at least partially) unvalidated EXIF-related metadata, yet we'll want to segregate it from the CSV-provided EXIF metadata.

msmolens:
Move metadata fields that have the "exif_" prefix to a new dedicated unstructured field. This separates EXIF metadata from other unstructured metadata.

Fixes #463

### DIFF
--- a/server/models/dataset_helpers/image_metadata.py
+++ b/server/models/dataset_helpers/image_metadata.py
@@ -858,6 +858,20 @@ def _checkMetadataWarnings(clinical):
     return warnings
 
 
+def _extractExifMetadata(data, unstructuredExif):
+    """
+    Add EXIF-related metadata to its own unstructured field.
+
+    :param data: The image metadata.
+    :type data: dict
+    :param unstructuredExif: The dictionary of unstructured EXIF metadata.
+    :type unstructuredExif: dict
+    """
+    for key in list(six.iterkeys(data)):
+        if key.lower().startswith('exif_'):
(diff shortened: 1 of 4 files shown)


server/models/dataset_helpers/image_metadata.py
```diff
@@ -858,6 +858,20 @@ def _checkMetadataWarnings(clinical):
...
+    """
+    Add EXIF-related metadata to its own unstructured field.
+
+    :param data: The image metadata.
+    :type data: dict
+    :param unstructuredExif: The dictionary of unstructured EXIF metadata.
+    :type unstructuredExif: dict
+    """
+    for key in list(six.iterkeys(data)):
+        if key.startswith('exif_'):
```
brianhelba: Let's do a case-insensitive comparison.

msmolens: The latest revision uses `lower()` in the comparison.

PR #568, opened 2018-05-21
Metadata API

msmolens:
Update the API to register, apply, and download metadata to avoid using Girder folder/file endpoints.

The `POST /dataset/{id}/metadata` endpoint now takes the metadata CSV data in the HTTP request body and the filename as a query parameter.

The `POST /dataset/{id}/metadata/{metadataFileId}` endpoint is unchanged except that the 'fileId' parameter is renamed to 'metadataFileId'.

The `GET /dataset/{id}/metadata/{metadataFileId}` endpoint is added to download a metadata file.

Tasks:
- [x] Update server
- [x] Update web client
- [x] Update tests


brianhelba:
Can we rename `POST /dataset/{id}/metadata/{metadataFileId}` -> `POST /dataset/{id}/metadata/{metadataFileId}/apply`?

brianhelba:
Can you also:
* Move files to be stored as `attachedTo` and get rid of the user folders
  * This requires a `DELETE /dataset/{id}/metadata/{metadataFileId}` endpoint for site admins to remove attached files, to ensure we have adequate options for file lifecycle management
* Require dataset `ADMIN` access to apply metadata to a dataset
  * There should then be no need to validate permissions on Files themselves

server/api/dataset.py
```diff
@@ -425,19 +425,14 @@ def _checkFileFormat(self, file, formats):
     )
     @access.cookie
     @access.public(scope=TokenScope.DATA_READ)
-    @loadmodel(model='file', map={'metadataFileId': 'metadataFile'}, force=True, exc=False)
+    @loadmodel(model='file', map={'metadataFileId': 'metadataFile'}, level=AccessType.READ,
+               exc=False)
```
msmolens: @brianhelba This line's change requires https://github.com/girder/girder/pull/2716. Although I think it's good defensive programming, I'm open to backing this change out since the access checks performed here are redundant: assuming metadata files will always attached to datasets, then the below access check on the dataset is sufficient.

msmolens:
@brianhelba PTAL, and see inline comment. Additionally, we should consider the impact on existing metadata files and whether any file/database migration is necessary.

server/api/dataset.py
```diff
@@ -425,19 +425,14 @@ def _checkFileFormat(self, file, formats):
     )
     @access.cookie
     @access.public(scope=TokenScope.DATA_READ)
-    @loadmodel(model='file', map={'metadataFileId': 'metadataFile'}, force=True, exc=False)
+    @loadmodel(model='file', map={'metadataFileId': 'metadataFile'}, level=AccessType.READ,
```
brianhelba: I think this should be `WRITE` access to the file (which is the same as `WRITE` access to the dataset) in order to download the raw CSV file. Normal users of the dataset (with `READ` access) have no need to access the CSV itself.

brianhelba: Also, can you add a comment, noting the fact that this is really `WRITE` access to the dataset? Since we don't use the permission checking of `attachedTo` files much, I don't want to forget where this permission comes from. Also, add a comment to other places in this file where do you do a similar thing.

brianhelba:
As part of the merge, I'll do a migration to add the `attachedTo` to existing files (and remove them from the folders where they live now).

PR #580, opened 2018-10-09
Remove upload credential settings

danlamanna:
Based on #577. This switches the data upload to use IAM roles instead of an explicitly passed user.


danlamanna:
@brianhelba This is ready for review. This is tested and working on staging.

danlamanna:
@brianhelba Tested working, ready for review

PR #584, opened 2018-10-19
Vagrant / ansible improvements

brianhelba:
* Don't hardcode Girder user in Ansible roles
* Use new list syntax for Ansible apt module
* Rename role "large_image"
* Rename development server to isic-archive.test, to work with Chrome

### DIFF
--- a/ansible/roles/isic/meta/main.yml
+++ b/ansible/roles/isic/meta/main.yml
@@ -4,7 +4,7 @@ dependencies:
   - role: girder
   - role: supervisor
   - role: worker
-  - role: large-image
+  - role: large_image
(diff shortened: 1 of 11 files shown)


ansible/roles/isic/meta/main.yml
```diff
@@ -4,7 +4,7 @@ dependencies:
   - role: girder
   - role: supervisor
   - role: worker
-  - role: large-image
+  - role: large_image
```
danlamanna: Any reason for this change other than preference?

brianhelba: Consistency with all the other naming of the large_image plugin, package, and repo.

PR #588, opened 2018-12-03
Don't require a user to be a member of a study to delete other users

brianhelba:


brianhelba:
Tests will fail until https://github.com/ImageMarkup/isic-archive/pull/587 is merged, but this was tested and works.

PR #593, opened 2019-01-29
Add asynchronous implementation of image ingestion

danlamanna:


brianhelba:
Tests are now passing on `master`. Can you drop the first commit (regarding Tox) and rebase?

isic-archive-tasks/isic_archive_tasks/image.py
```diff
@@ -0,0 +1,199 @@
...
+    if image['ingested']:
+        logger.warning('Attempting to ingest an already ingested image.')
+        return
+
+    try:
+        imageData = Image().imageData(image)
+        image['meta']['acquisition']['pixelsY'] = imageData.shape[0]
+        image['meta']['acquisition']['pixelsX'] = imageData.shape[1]
+        image['readable'] = True
+        image['ingested'] = False
```
brianhelba: This should already be `False` from when the image was created.

isic-archive-tasks/isic_archive_tasks/image.py
```diff
@@ -0,0 +1,199 @@
...
+        image['ingestionState'] = {
+            'large-image': None,
```
brianhelba: Most of our "schema" uses camelCase for keys.

isic-archive-tasks/isic_archive_tasks/image.py
```diff
@@ -0,0 +1,199 @@
...
+def ingestImage(imageId):
+    from girder.plugins.isic_archive.models.image import Image
+    image = Image().load(imageId, force=True)
+
+    if image['ingested']:
+        logger.warning('Attempting to ingest an already ingested image.')
+        return
+
+    try:
+        imageData = Image().imageData(image)
```
brianhelba: Can we only wrap these first 3 lines in the `try` block? They are the only ones that are expected to sometimes fail with non-image files.

isic-archive-tasks/isic_archive_tasks/image.py
```diff
@@ -0,0 +1,199 @@
...
+            'large-image': None,
+            'superpixel-mask': None
+        }
+        Image().save(image)
+
+        callback = markImageIngested.s(image['_id'])
+        chord([generateSuperpixels.s(image['_id']),
+               generateLargeImage.s(image['_id'])])(callback)
+    except Exception:
+        sentry_sdk.capture_exception()
```
brianhelba: Probably don't need this.

isic-archive-tasks/isic_archive_tasks/image.py
```diff
@@ -0,0 +1,199 @@
...
+        image['readable'] = False
```
brianhelba: Can you set `'readable'` to `False` or `None` when the image document is first created?

isic-archive-tasks/isic_archive_tasks/image.py
```diff
@@ -0,0 +1,199 @@
...
+            'size': len(superpixelsEncodedStream.getvalue())
+        }, data=superpixelsEncodedStream.getvalue(), headers={
+            'Girder-Token': self.token
+        })
+        uploadSuperpixelsResponse.raise_for_status()
+
+        superpixelsFile = File().load(uploadSuperpixelsResponse.json()['_id'], force=True)
+        image = Image().load(imageId, force=True, exc=False)
+
+        superpixelsFileNameMatch = re.match(
```
brianhelba: Remove this match and conditional. It was left over from the event handler, which was called for every file upload, and we needed to filter non-superpixel files.

isic-archive-tasks/isic_archive_tasks/image.py
```diff
@@ -0,0 +1,199 @@
...
+
+        superpixelsData = ScikitSegmentationHelper.superpixels(originalImageData)
+        superpixelsEncodedStream = ScikitSegmentationHelper.writeImage(
+            superpixelsData, 'png')
+
+        uploadSuperpixelsResponse = requests.post('http://isic-archive.test/api/v1/file', params={
+            'parentType': 'item',
+            'parentId': imageId,
+            'name': '%s_superpixels_v%s.png' % (image['name'], SUPERPIXEL_VERSION),
+            'size': len(superpixelsEncodedStream.getvalue())
```
brianhelba: Need to set `mimeType` here because the server doesn't auto-detect.

isic-archive-tasks/isic_archive_tasks/image.py
```diff
@@ -0,0 +1,199 @@
...
+        superpixelsFile = File().load(uploadSuperpixelsResponse.json()['_id'], force=True)
+        image = Image().load(imageId, force=True, exc=False)
+
+        superpixelsFileNameMatch = re.match(
+            r'^%s_superpixels_v([0-9.]+)\.png' % image['name'],
+            superpixelsFile['name'])
+        if not superpixelsFileNameMatch:
+            return
+
+        superpixelsVersion = float(superpixelsFileNameMatch.group(1))
```
brianhelba: Version is known from the global, don't parse it.

isic-archive-tasks/isic_archive_tasks/image.py
```diff
@@ -0,0 +1,199 @@
...
+        superpixelsFile['superpixelVersion'] = superpixelsVersion
+        # Work around an upstream Girder bug where "File().validate" sets
+        #   superpixelsFile['exts'] = ['0', 'png']
+        superpixelsFile = File().validate(superpixelsFile)
```
brianhelba: I'm not sure why we need to validate again, we can discuss more if you can't figure it out, but probably remove.

isic-archive-tasks/isic_archive_tasks/image.py
```diff
@@ -0,0 +1,199 @@
...
+                vips_tiffsave(input_file.name, output_file.name)
+
+                newFilename = os.path.splitext(imageFile['name'])[0] + time.strftime(
+                    '%Y%m%d-%H%M%S') + '.tiff'
+                uploadLargeImageResp = requests.post('http://isic-archive.test/api/v1/file',
+                                                     params={
+                                                         'parentType': 'item',
+                                                         'parentId': imageId,
+                                                         'name': newFilename,
+                                                         'size': os.path.getsize(
```
brianhelba: Need to set `mimeType`.

isic-archive-tasks/isic_archive_tasks/image.py
```diff
@@ -0,0 +1,199 @@
...
+
+        superpixelsFileNameMatch = re.match(
+            r'^%s_superpixels_v([0-9.]+)\.png' % image['name'],
+            superpixelsFile['name'])
+        if not superpixelsFileNameMatch:
+            return
+
+        superpixelsVersion = float(superpixelsFileNameMatch.group(1))
+        superpixelsFile['superpixelVersion'] = superpixelsVersion
+        # Work around an upstream Girder bug where "File().validate" sets
```
brianhelba: Leave this comment, as it applies to why we set `validate=False` on the `save`

isic-archive-tasks/isic_archive_tasks/image.py
```diff
@@ -0,0 +1,199 @@
...
+        #   superpixelsFile['exts'] = ['0', 'png']
+        superpixelsFile = File().validate(superpixelsFile)
+        superpixelsFile['exts'] = ['png']
+        superpixelsFile = File().save(superpixelsFile, validate=False)
+
+        image['superpixelsId'] = superpixelsFile['_id']
+        image['ingestionState']['superpixel-mask'] = True
+        Image().save(image)
+    except Exception:
+        sentry_sdk.capture_exception()
```
brianhelba: Don't need?

isic-archive-tasks/isic_archive_tasks/image.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+@app.task(bind=True, base=CredentialedGirderTask)
+def generateLargeImage(self, imageId):
+    from girder.plugins.isic_archive.models.image import Image
+    from girder.plugins.large_image.models.image_item import ImageItem
+    try:
+        imageItem = ImageItem().load(id=imageId, force=True)
+        imageFile = Image().originalFile(imageItem)
+        imageItem['largeImage'] = {
```
brianhelba: Don't do this initially, since we don't want it if the creation fails.

isic-archive-tasks/isic_archive_tasks/image.py
```diff
@@ -0,0 +1,199 @@
...
+            'http://isic-archive.test/api/v1/file/%s/download' % imageFile['_id'], headers={
+                'Girder-Token': self.token
+            }, stream=True)
+        originalFileStreamResponse.raise_for_status()
+
+        with tempfile.NamedTemporaryFile() as input_file:
+            input_file.write(originalFileStreamResponse.content)
+            input_file.flush()  # necessary
+
+            with tempfile.NamedTemporaryFile() as output_file:
```
brianhelba: Can you combine the `with`: https://stackoverflow.com/questions/893333/multiple-variables-in-a-with-statement

isic-archive-tasks/isic_archive_tasks/image.py
```diff
@@ -0,0 +1,199 @@
...
+                vips_tiffsave(input_file.name, output_file.name)
+
+                newFilename = os.path.splitext(imageFile['name'])[0] + time.strftime(
```
brianhelba: The name should be `imageItem['name'] + '.tiff'`; e.g. `ISIC_1234567.tiff`

isic-archive-tasks/isic_archive_tasks/image.py
```diff
@@ -0,0 +1,199 @@
...
+                    '%Y%m%d-%H%M%S') + '.tiff'
+                uploadLargeImageResp = requests.post('http://isic-archive.test/api/v1/file',
+                                                     params={
```
brianhelba: Can you fix the formatting? Drop the first argument down to a new indentation level, and then you can move everything else left.

isic-archive-tasks/isic_archive_tasks/image.py
```diff
@@ -0,0 +1,199 @@
...
+                                                         'parentType': 'item',
+                                                         'parentId': imageId,
+                                                         'name': newFilename,
+                                                         'size': os.path.getsize(
+                                                             output_file.name)
+                                                     }, data=open(output_file.name).read(),
```
brianhelba: A context manager is probably safer to use here when opening `output_file.name` for reading.

isic-archive-tasks/isic_archive_tasks/image.py
```diff
@@ -0,0 +1,199 @@
...
+
+    logger.info('running "%s"' % ' '.join(convert_command))
+    proc = subprocess.Popen(convert_command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    out, err = proc.communicate()
+
+    if out.strip():
+        logger.info('stdout: ' + out)
+
+    if err.strip():
+        logger.info('stderr: ' + err)
```
brianhelba: This should probably be a higher log level than `info`.

isic-archive-tasks/isic_archive_tasks/image.py
```diff
@@ -0,0 +1,199 @@
...
+
+    if proc.returncode:
+        raise Exception('VIPS command failed (rc=%d): %s' % (proc.returncode,
```
brianhelba: We already log the convert command, do you need it again?

isic-archive-tasks/isic_archive_tasks/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+register('jsonpickle', jsonpickle.encode, jsonpickle.decode, content_type='application/json',
+         content_encoding='utf-8')
+
+
+@app.task()
+def sendIngestionNotifications():
+    from girder.plugins.isic_archive.models.batch import Batch
+    from girder.plugins.isic_archive.models.image import Image
+    for batch in Batch().find({'ingestStatus': 'extracted'}):
+        if Batch().hasImagesPendingIngest(batch):
```
brianhelba: You could invert the conditional if you want and get rid of `return`

isic-archive-tasks/isic_archive_tasks/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+            return
+        else:
+            failedImages = list(Image().find({'$and': [
```
brianhelba: `'$and'` should not be necessary

server/mail_templates/ingestDatasetConfirmation.mako
```diff
@@ -51,6 +51,14 @@ Thank you for contributing a ZIP image batch to the ISIC Archive.
     </tbody>
 </table>
 
+% if failedImages:
+    <h4>Failures</h4>
+    <br>One or more files within your dataset failed to be completely imported into the archive:<br>
```
brianhelba: @brianhelba Review copy and clean up

isic-archive-tasks/isic_archive_tasks/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+                {'meta.batchId': batch['_id']},
+                {'$or': [
+                    {'ingestionState.large-image': False},
+                    {'ingestionState.superpixel-mask': False}]}]}))
+
```
brianhelba: Can we run a second query, looking for images with `'readable': False`? We should add this to the email as a list of "skipped files".

isic-archive-tasks/isic_archive_tasks/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+        super(CredentialedGirderTask, self).__call__(*args, **kwargs)
+
+
+register('jsonpickle', jsonpickle.encode, jsonpickle.decode, content_type='application/json',
+         content_encoding='utf-8')
+
+
+@app.task()
+def sendIngestionNotifications():
```
brianhelba: Feel free to rename to something that indicates this function may do no actual work.

isic-archive-tasks/isic_archive_tasks/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+            Batch().save(batch)
+
+
+@app.task(bind=True, base=CredentialedGirderTask)
+def sendIngestionNotification(self, batchId, datasetId, userId, failedImages):
+    from girder.plugins.isic_archive.models.batch import Batch
+    from girder.plugins.isic_archive.models.dataset import Dataset
+    from girder.plugins.isic_archive.utility.mail_utils import sendEmail, sendEmailToGroup
+    batch = Batch().load(batchId)
+    dataset = Dataset().load(datasetId, force=True)
```
brianhelba: `dataset` and `user` should come out of `batch`, and not be parameters

server/mail_templates/ingestDatasetConfirmation.mako
```diff
@@ -51,6 +51,14 @@ Thank you for contributing a ZIP image batch to the ISIC Archive.
...
+    % for failedImage in failedImages:
+        <a href="${host}/girder#item/${failedImage['_id']}">${failedImage['name']}</a><br>
```
brianhelba: We shouldn't link to any internal Girder pages. Plain text with no link is fine for the list content.

brianhelba: Use `failedImage['privateMeta']['originalFilename']` for the name. The `'name'` like `ISIC_1234567` is not meaningful in this context

isic-archive-tasks/isic_archive_tasks/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+    user = User().load(userId, force=True)
+    host = mail_utils.getEmailUrlPrefix()
+    params = {
+        'group': False,
```
brianhelba: Assuming this isn't called anywhere else, rename `'group'` to something like `'originalUploader'` (here and in the template). This doesn't make sense now.

isic-archive-tasks/isic_archive_tasks/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+        'host': host,
+        'dataset': dataset,
+        # We intentionally leak full user details here, even though all
+        # email recipients may not have access permissions to the user
+        'user': user,
+        'batch': batch,
+        'failedImages': failedImages
+    }
+    subject = 'ISIC Archive: Dataset Upload Confirmation'
```
brianhelba: Maybe change the subject depending on whether we have any `failedImages`

isic-archive-tasks/isic_archive_tasks/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+                    {'ingestionState.superpixel-mask': False}]}]}))
+
+            sendIngestionNotification.delay(batch['_id'], batch['datasetId'],
+                                            batch['creatorId'], failedImages)
+            batch['ingestStatus'] = 'notified'
+            Batch().save(batch)
+
+
+@app.task(bind=True, base=CredentialedGirderTask)
+def sendIngestionNotification(self, batchId, datasetId, userId, failedImages):
```
brianhelba: Don't need the `self` parameter

isic-archive-tasks/isic_archive_tasks/zip.py
```diff
@@ -0,0 +1,126 @@
...
+
+        # Remove upload information from batch
+        del batch['s3BucketName']
+        del batch['s3ObjectKey']
+        batch['ingestStatus'] = 'extracting'
+        batch = Batch().save(batch)
+
+        # Process zip file
+        with ZipFileOpener(fileName) as (fileList, fileCount):
+            for originalFilePath, originalFileRelpath in fileList:
```
brianhelba: We need to add a policy where duplicate **basenames** of files in a ZIP are not allowed (rejecting the ZIP). Also, if any basename of a file in a ZIP matches the `privateMeta.originalFilename` of an image already saved in the dataset, the ZIP should be rejected.

server/mail_templates/ingestDatasetConfirmation.mako
```diff
@@ -51,6 +51,14 @@ Thank you for contributing a ZIP image batch to the ISIC Archive.
     </tbody>
 </table>
 
+% if failedImages:
+    <h4>Failures</h4>
+    <br>One or more files within your dataset failed to be completely imported into the archive:<br>
+    % for failedImage in failedImages:
```
brianhelba: If this list is excessively long, we should truncate it. However, that may be too much work for this PR. Can you add a TODO comment and/or file an issue?

isic-archive-tasks/isic_archive_tasks/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+            return
+        else:
+            failedImages = list(Image().find({'$and': [
+                {'meta.batchId': batch['_id']},
+                {'$or': [
+                    {'ingestionState.large-image': False},
+                    {'ingestionState.superpixel-mask': False}]}]}))
+
+            sendIngestionNotification.delay(batch['_id'], batch['datasetId'],
+                                            batch['creatorId'], failedImages)
```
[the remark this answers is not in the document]
danlamanna: Max message size is actually 2GB, even though messages are ideally very small. This shouldn't be a problem.

isic-archive-tasks/isic_archive_tasks/zip.py
danlamanna: I'm going to tackle this in a follow up PR.

danlamanna:
@brianhelba This is ready for another pass. There are a few things left to resolve (outside of the outstanding comments on this PR):
- URL setting
- MongoClient is complaining about not being fork safe, I think this means I need to run `fixGirderImports` at a later time.
- I left the `$and` syntax, there was some logic in `_findQueryFilter` that expects a dictionary rather than a list, and I wasn't sure what the best way around that was.
- The large image creation is incorrect, I'm still getting read permission errors when trying to access the viewer.
- I can still get this to break in odd ways if I put exceptions in various places. Such that a success email will go out even when things have catastrophically failed. I have to look into this a bit.

danlamanna:
I also created a new branch based off this for refactoring the ansible/vagrant: https://github.com/ImageMarkup/isic-archive/compare/celery-upload...celery-upload-infra.

danlamanna:
- Add email for individual image upload

danlamanna:
An additional step for future hardening:
- Add very short connection timeouts to all requests: boto, requests, smtp server, etc.

PR #595, opened 2019-01-31
Handle an API change with OpenCV 4

brianhelba:
See https://github.com/opencv/opencv/pull/12663


brianhelba:
These tests now pass. Failing style tests will be fixed separately.

PR #606, opened 2019-02-19
Remove unnecessary lines from the Vagrantfile

brianhelba:


danlamanna:
Did the bionic box stop leaving that file?

PR #609, opened 2019-02-19
Set cookies for all subdomains

brianhelba:


brianhelba:
This has a number of negative impacts, where other domains (`staging.`, `submission.challenge.`, etc.) will overwrite logins to the apex domain, but it fixes the most critical issues with `www`<->apex interoperability, which impacts Discourse SSO and use of the annotation tool.

PR #614, opened 2019-03-22
 Replace supervisor in favor of systemd

danlamanna:


ansible/roles/isic/tasks/main.yml
```diff
@@ -51,6 +51,24 @@
...
 
+- name: Deploy Girder systemd service
+  template:
+    src: girder.service.j2
+    dest: "/etc/systemd/system/girder.service"
+    owner: root
+    group: root
+    mode: 0644
+  become: yes
+  notify: Restart Girder
```
brianhelba: This is the only place where the `Restart Girder` handler actually needs `daemon_reload: true`, right? If so, could we make a second handler to `Reload Girder systemd service` (or something), which just runs `daemon_reload`?

It's not that `daemon_reload` is *that* expensive, but it'd be cleaner to avoid calling it for routine restarts due to app updates. Or maybe I'm totally misunderstanding `daemon_reload`....

brianhelba: Let's merge this as-in. We can try this change on another PR, which would cover other services too.

PR #619, opened 2019-03-22
Use multipart upload for image ingestion

danlamanna:


isic_archive/tasks/zip.py
```diff
@@ -38,11 +38,25 @@ def _uploadZipfileToGirder(requestSession, filePath, dataset):
...
                 'mimeType': 'application/zip'
             },
-            data=fileStream,
         )
         uploadFileResponse.raise_for_status()
 
-    uploadFile = File().load(uploadFileResponse.json()['_id'], force=True)
+        chunk_size = 1024 * 1024 * 50
+        offset = 0
+        while True:
```
brianhelba: https://docs.python.org/3/library/functions.html#iter ?

isic_archive/tasks/zip.py
```diff
@@ -38,11 +38,25 @@ def _uploadZipfileToGirder(requestSession, filePath, dataset):
...
+            chunk = fileStream.read(min(chunk_size, (fileSize - offset)))
+
+            if not chunk:
+                break
+
+            r = requestSession.post('file/chunk',
+                                    params={'offset': offset,
+                                            'uploadId': uploadFileResponse.json()['_id']},
```
brianhelba: Pretty minor, but we don't need to re-parse the JSON in every iteration. I'd grab the ID once in a variable above.

isic_archive/tasks/zip.py
```diff
@@ -38,11 +38,25 @@ def _uploadZipfileToGirder(requestSession, filePath, dataset):
...
-    uploadFile = File().load(uploadFileResponse.json()['_id'], force=True)
+        chunk_size = 1024 * 1024 * 50
+        offset = 0
+        while True:
+            chunk = fileStream.read(min(chunk_size, (fileSize - offset)))
+
+            if not chunk:
+                break
+
+            r = requestSession.post('file/chunk',
```
brianhelba: Can you use a better name, like `uploadChunkResponse`?

PR #622, opened 2019-03-24
Python 3

danlamanna:


.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2
 jobs:
   serverTest:
     docker:
-      - image: isic/isic_test:latest
+      - image: isic/isic_test:latest-py3
```
danlamanna: Remove this change before merge.

danlamanna:
I found some bugs with this branch, it's not ready for merging yet.

danlamanna:
@brianhelba good to go?

isic_archive/models/dataset.py
```diff
@@ -580,14 +581,16 @@ def applyMetadata(self, dataset, metadataFile, save):
...
+        for chunk in File().download(metadataFile, headers=False)():
+            metadataFileStream.write(chunk)
 
         images = []
         metadataErrors = []
         metadataWarnings = set()
 
         try:
-            csvReader = csv.DictReader(generateLines(metadataFileStream))
+            csvReader = csv.DictReader(metadataFileStream.getvalue().decode('utf-8').splitlines())
```
brianhelba: I think we should use `'utf-8-sig'` here. If the file includes a BOM at the beginning, it will properly decode it; see: https://docs.python.org/3/library/codecs.html#encodings-and-unicode If the file doesn't include a BOM, it should behave the same as `'utf-8'`: https://stackoverflow.com/a/44573867

This is something we could easily write a test for, by generating a sample CSV file with a BOM and without.

PR #631, opened 2019-04-02
Remove six

brianhelba:


isic_archive/utility/mail_utils.py
```diff
@@ -28,13 +26,10 @@
 
 
 def sendEmail(to, subject, text):
-    if isinstance(to, six.string_types):
+    if isinstance(to, str):
         to = (to,)
 
-    if isinstance(text, six.text_type):
```
brianhelba: This is the only non-trivial change. I can't tell from [the `MimeText` docs](https://docs.python.org/3/library/email.mime.html#email.mime.text.MIMEText) what the correct type of `text` here should be, particularly since we pass `'UTF-8'`. Some actual testing is needed.

brianhelba: The source code of `MimeText` and https://docs.python.org/3.4/library/email-examples.html indicate that the first argument should be a unicode string in both Python 2 and 3. This is likely incorrect in upstream Girder too, though the `MimeText` implementation is somewhat tolerant of passing `bytes` too.

PR #634, opened 2019-04-08
Use the 'pil' IO plugin for scikit-image

brianhelba:
The default plugin changed to 'imageio' in a recent release of scikit-image.


brianhelba:
`scikit-image` 0.15 changed the default plugin: http://scikit-image.org/docs/dev/release_notes_and_installation.html#api-changes

The new `imageio` plugin uses `format` instead of `format_str`.

Improved test coverage or version pinning would have caught or prevented this.

PR #638, opened 2019-04-08
Ansible format improvements

brianhelba:


ansible/.yamllint
```diff
@@ -0,0 +1,11 @@
+---
+extends: default
+
+ignore: |
+  node_modules
```
brianhelba: Need to update this.

ansible/roles/minio/tasks/main.yml
```diff
@@ -32,10 +32,20 @@
     timeout: 20
 
 - name: Create default bucket
-  shell: |
-    . {{ python_dist_path }}/bin/activate
-    export AWS_ACCESS_KEY_ID=accesskey
-    export AWS_SECRET_ACCESS_KEY=secretkey
-    {{ python_dist_path }}/bin/aws --endpoint-url http://localhost:9000 s3api create-bucket --bucket {{ isic_upload_bucket_name }} && touch {{ ansible_user_dir }}/.bucket.created
+  command: >-
```
brianhelba: @danlamanna I refactored this command more heavily. Can you review.

danlamanna: It turns out minio treats all directories within the top level datadir as buckets. So this while thing can just be simplified to creating `isic_upload_bucket_name`.

brianhelba: So something like `creates: {{ isic_upload_bucket_name }}`? We should make this a fully qualified path, so where do the buckets live?

danlamanna: I fixed this.

PR #643, opened 2019-04-10
Remove unnecessary mongo flag for CI

danlamanna:
CircleCI did some fixing of their mongo images and caused this flag to no longer be necessary.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ jobs:
     docker:
       - image: isic/isic_test:latest
       - image: circleci/mongo:3.6-ram
-        command: ["mongod", "--storageEngine", "ephemeralForTest", "--dbpath", "/dev/shm/mongo"]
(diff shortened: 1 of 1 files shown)


brianhelba:
Can you provide more explanation?

danlamanna:
Taking another look at this, I'm pretty thoroughly confused. The master build started breaking last night, right after https://github.com/circleci/circleci-images/commit/93920c60a11c41a074957904d9e0b69badcd9dba was pushed. It looks like our config should have continued to work, but it failed because /dev/shm/mongo wasn't a directory.

Now I'm wondering if Circle was caching the old docker image and that was what was causing the failure?

brianhelba:
I thought I fixed that already: https://github.com/circleci/circleci-images/commit/b1d792c1b998c53c66b1886c367aa6348b9e64a6#diff-7221bde7d5c5713668228adf7cb17b9f

I'll take a look later.

PR #644, opened 2019-04-10
Add a script for annotation recovery

danlamanna:


danlamanna:
@brianhelba What should we do with this PR?

brianhelba:
Let's close. There's nothing valuable enough to want to archive in the repo.

Issue #649, opened 2019-06-21
girder restart issue


jettmcp:
Ansible is failing at the girder restart after mongodb restart and running handler [girder : wait_for]. This times out.

Issue seems to be with the mail_utils.py where traceback error is 
from girder.constants import 'SettingKey'
ImportError: cannot import name 'SettingKey'

I can't get past this.

jettmcp:
Thanks - that fixed that problem. I got past that but then crashed at Create asset store.

Ansible output:
fatal: [web]: FAILED! => changed=false 
  msg: |-
    <class 'girder_client.HttpError'>:HTTP error 403: GET http://localhost:8080/api/v1/assetstore
    Response text: {"message": "Administrator access required.", "type": "access"}
    {"message": "Administrator access required.", "type": "access"}
    Traceback (most recent call last):
      File "/tmp/ansible_girder_payload_8_dfgn6q/__main__.py", line 1831, in main
        gcm(module)
      File "/tmp/ansible_girder_payload_8_dfgn6q/__main__.py", line 1106, in __call__
        self.__process(method)
      File "/tmp/ansible_girder_payload_8_dfgn6q/__main__.py", line 1134, in __process
        ret = getattr(self, method)(*args, **kwargs)
      File "/tmp/ansible_girder_payload_8_dfgn6q/__main__.py", line 1663, in assetstore
        assetstores = {a['name']: a for a in self.get("assetstore")}
      File "/usr/local/lib/python3.6/dist-packages/girder_client/__init__.py", line 503, in get
        return self.sendRestRequest('GET', path, parameters, jsonResp=jsonResp)
      File "/usr/local/lib/python3.6/dist-packages/girder_client/__init__.py", line 497, in sendRestRequest
        response=result)
    girder_client.HttpError: HTTP error 403: GET http://localhost:8080/api/v1/assetstore
    Response text: {"message": "Administrator access required.", "type": "access"}

PLAY RECAP *********************************************************************
web                        : ok=50   changed=41   unreachable=0    failed=1    skipped=2    rescued=0    ignored=0   

Monday 24 June 2019  19:23:32 +0000 (0:00:00.724)       0:10:39.940 *********** 
=============================================================================== 
girder : command ------------------------------------------------------- 97.42s
large_image : Install vips --------------------------------------------- 75.91s
isic : Build ISIC Admin GUI -------------------------------------------- 75.83s
isic : Install ISIC GUI dependencies ----------------------------------- 73.83s
[...]

PR #650, opened 2019-06-24
Fix Girder 3 compatibility issues

danlamanna:
@brianhelba


danlamanna:
This should fix #649.

PR #664, opened 2019-07-29
Fix batch upload widget to work with S3

brianhelba:


isic-archive-gui/src/Datasets/UploadBatch.vue
```diff
@@ -0,0 +1,198 @@
...
+        },
+      });
+
+      const s3Options = {
+        apiVersion: '2006-03-01',
+        accessKeyId: initUpload.accessKeyId,
+        secretAccessKey: initUpload.secretAccessKey,
+        sessionToken: initUpload.sessionToken,
+      };
+      if (process.env.NODE_ENV !== 'production') {
```
brianhelba: @danlamanna To work in a Vagrant + Minio environment, the web client needs to know to use some different options. Would it be reasonable to pass some of this information (or at least provide some sort of `devMode = true`-type) via the upload initialize API?

PR #668, opened 2019-08-07
Use refactored mail_utils from upstream Girder

brianhelba:
This incorporates the upstream changes from https://github.com/girder/girder/pull/3103 .


danlamanna:
Is this blocked? It's breaking #669.

PR #671, opened 2019-08-28
Add feature to reingest a batch

danlamanna:


danlamanna:
Updated infra wii: https://github.com/ImageMarkup/isic-infrastructure/wiki/Common-Problems#reingesting-an-entire-batch

PR #688, opened 2020-01-10
Make pip install under Ansible idempotent

brianhelba:


brianhelba:
This won't work until we are able to apply versioning to each publication of `isic-archive`. Otherwise, there's no way to tell whether the Python package has changed.

PR #693, opened 2020-01-21
Fix systemd environment variable issues

brianhelba:


ansible/roles/isic/templates/services/celerybeat.service.j2
```diff
@@ -9,12 +9,15 @@ StateDirectory=celerybeat
 RuntimeDirectory=celerybeat
 Environment=LC_ALL=C.UTF-8
 Environment=LANG=C.UTF-8
-ExecStart=/bin/sh -c '{{ girder_virtualenv }}/bin/celery \
+# TODO: systemd<240 does not provide STATE_DIRECTORY and RUNTIME_DIRECTORY automatically
```
brianhelba: @danlamanna FYI, this includes 18.04. Anything >= 19.04 has the new variables.

PR #695, opened 2020-01-22
Do not allow redirects when making HTTP requests in tasks

brianhelba:
This can mask misconfigurations, particularly when a 301 causes a POST to change to a GET.

Also, use only the base uploaded file name for temporary uploads


brianhelba:
https://github.com/psf/requests/blob/9b95b4e7b4df1865b058161f1492989868614535/requests/sessions.py#L329

PR #699, opened 2020-01-25
Refactor MailHog Ansible role

brianhelba:


ansible/roles/mailhog/templates/mailhog.service.j2
```diff
@@ -1,22 +1,13 @@
...
-Environment=MH_UI_BIND_ADDR=0.0.0.0:8025
-Environment=MH_SMTP_BIND_ADDR=0.0.0.0:1025
-Environment=MH_STORAGE=maildir
-Environment=MH_MAILDIR_PATH={{ ansible_user_dir }}/mailhog-storage
-{% if mailhog_username is defined and mailhog_password_bcrypt is defined %}
-Environment=MH_AUTH_FILE=/etc/mailhog-auth
-{% endif %}
+EnvironmentFile=/etc/mailhog.env
 ExecStart=/usr/local/bin/mailhog
-StandardOutput=journal
```
brianhelba: @danlamanna FYI, `StandardOutput=journal` is the default for all services.

ansible/roles/mailhog/tasks/main.yml
```diff
@@ -7,26 +7,37 @@
...
     mode: 0755
   become: true
+  notify:
+    - Restart MailHog
 
 - name: Add MailHog basic authentication file
   copy:
-    content: |
-      {{ mailhog_username }}:{{ mailhog_password_bcrypt }}
+    content: "{{ mailhog_username }}:{{ mailhog_password|password_hash('bcrypt') }}"
```
brianhelba: @danlamanna Ansible can generate hashes: https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#hashing-filters

ansible/roles/mailhog/handlers/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: Restart MailHog
```
brianhelba: The handler ensures that future changes to the config file will restart the service.

PR #710, opened 2020-04-06
Make application of metadata asynchronous

danlamanna:
----


danlamanna:
@brianhelba The ability to have linting on isic-archive-gui/src/Datasets/ApplyMetadataView.js would help. `yarn run lint` yields no errors even when some exist.

PR #712, opened 2020-04-06
Add original filename of zip to batch

danlamanna:
Note, this only works on our client.

### DIFF
--- a/isic_archive/api/dataset.py
+++ b/isic_archive/api/dataset.py
@@ -322,6 +322,7 @@ class DatasetResource(IsicResource):
                )
         .param('id', 'The ID of the dataset.', paramType='path')
         .param('signature', 'Signature of license agreement.', paramType='form')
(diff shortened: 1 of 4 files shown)


isic_archive/api/dataset.py
```diff
@@ -337,7 +338,7 @@ def initiateZipUploadToS3(self, dataset, params):
             raise ValidationException('Signature must be specified.', 'signature')
 
         try:
-            return Dataset().initiateZipUploadS3(dataset=dataset, signature=signature, user=user)
+            return Dataset().initiateZipUploadS3(dataset=dataset, signature=signature, user=user, filename=params.get('filename', None))
```
brianhelba, line 341: Could you extract `filename` as a separate variable? It may make tracing errors from Sentry easier.

danlamanna: Fixed. Though FWIW this wouldn't help with Sentry since params is in the local scope.

PR #713, opened 2020-04-06
Give scikit-image its own image handle

danlamanna:
scikit-image will close the file handle it's given as seen
[here](https://github.com/scikit-image/scikit-image/blob/4ed2079ea40ee2ed47d80552a316ea707e9ceef9/skimage/io/_plugins/pil_plugin.py#L125).
This isn't new, though it only occurs on tiff images.


danlamanna:
Fixes https://sentry.io/organizations/isic/issues/1590854192/?project=1378210&query=is%3Aunresolved

PR #716, opened 2020-04-09
Add random generation of ISIC ids

danlamanna:


isic_archive/models/image.py
```diff
@@ -29,16 +28,26 @@ def initialize(self):
...
+            taken = bool(Image().findOne({'name': f'ISIC_{id_}'}).count())
+
+        return f'ISIC_{id_}'
+
     def createEmptyImage(self, originalFileRelpath, parentFolder, creator, dataset,
                          batch):
-        newIsicId = Setting().get(PluginSettings.MAX_ISIC_ID) + 1
+        newIsicId = self.generateIsicID()
         image = self.createItem(
             name='ISIC_%07d' % newIsicId,
```
brianhelba, line 46: Need to actually use the new value:

```suggestion
            name=newIsicId,
```

isic_archive/models/image.py
```diff
@@ -29,16 +28,26 @@ def initialize(self):
...
+    @staticmethod
+    def generateIsicID():
+        taken = True
+        id_: int = 0
+
+        while taken:
+            id_ = secrets.randbelow(10_000_000)
+            taken = bool(Image().findOne({'name': f'ISIC_{id_}'}).count())
+
+        return f'ISIC_{id_}'
```
brianhelba, line 40: * Need to string-format the new value to ensure leading zeros.

* `findOne` returns a `dict` or `None`, so you need to use `find` to work with `.count()`. In general, `.find().count()` is preferred to `.findOne()`, since the actual document is not returned.

```suggestion
			isicId = f'ISIC_{id_:07d}'
            taken = Image().find({'name': isicId}).count() > 0

        return isicId
```

isic_archive/models/image.py
```diff
@@ -29,16 +28,26 @@ def initialize(self):
 
         self.prefixSearchFields = ['lowerName', 'name']
 
+    @staticmethod
+    def generateIsicID():
+        taken = True
+        id_: int = 0
+
+        while taken:
+            id_ = secrets.randbelow(10_000_000)
```
brianhelba, line 37: If you think it makes it more obvious, we could encode the max ID using the same "7" digit places that the string-format uses, but it's up to you.

```suggestion
            id_ = secrets.randbelow(10**7)
```

isic_archive/models/image.py
```diff
@@ -29,16 +28,26 @@ def initialize(self):
 
         self.prefixSearchFields = ['lowerName', 'name']
 
+    @staticmethod
+    def generateIsicID():
```
brianhelba, line 32: Do we have the infrastructure to easily add a test, to assert that this returns something like `ISIC_\d{7}`? If this will take more than 10 minutes, I don't think it's worth it.

PR #719, opened 2020-04-13
Add method for resetting an image

danlamanna:


isic_archive/models/image.py
```diff
@@ -68,6 +78,29 @@ def createEmptyImage(self, originalFileRelpath, parentFolder, creator, dataset,
 
         return image
 
+    def resetImageForIngest(self, image):
+        originalFile = self.originalFile(image)
+        keepImageId = originalFile['_id'] if originalFile else None
```
brianhelba, line 83: How about just calling this `originalFileId`?

isic_archive/models/image.py
```diff
@@ -68,6 +78,29 @@ def createEmptyImage(self, originalFileRelpath, parentFolder, creator, dataset,
...
+        attrsToDelete = [
+            'ingested',
+            'readable',
+            'ingestionState',
+            'largeImage',
+            'superpixelsId',
+        ]
+
+        for file_ in self.childFiles(image):
```
brianhelba, line 92: Maybe call this `childFile`, so we don't get further into name style issues.

isic_archive/models/image.py
```diff
@@ -68,6 +78,29 @@ def createEmptyImage(self, originalFileRelpath, parentFolder, creator, dataset,
 
         return image
 
+    def resetImageForIngest(self, image):
+        originalFile = self.originalFile(image)
```
brianhelba, line 82: I'm concerned about the fact that `Image.originalFile` claims it will sometimes return the wrong file for old images with no large_image.

Since we can't rely on large_image to always provide the original file, I think we should move towards being more explicit in how we identify each type of image file.

We could either set:
1. `image['originalFileId'] = originalFile['_id']`
2. `file['imageRole'] = 'original'`

and so on for `superpixel`, `large_image`, etc. as a much more direct and reliable way to reference these files.

I think the major difference in the approaches is that 2. theoretically allows for multiple files in the same role. Although my first instinct is to prevent this with 1., I realize that it doesn't provide a truly strict constraint, as files could always leak if a `originalFileId` field were re-set without deleting the old file. The advantage of 2. is that it never leaks files, and it'd be much simpler to do a health check by asserting that each image has exactly one file of each role.

This whole thing is another side-task, but perhaps its worth adding now, to give more sanity and assurance to any maintenance going forward.

isic_archive/models/image.py
```diff
@@ -68,6 +78,29 @@ def createEmptyImage(self, originalFileRelpath, parentFolder, creator, dataset,
...
+        keepImageId = originalFile['_id'] if originalFile else None
+        attrsToDelete = [
```
brianhelba, line 84: It looks like for images created along the normal pipeline, these fields are initialized in varying ways:
* `ingested` [is set to `False`](https://github.com/ImageMarkup/isic-archive/blob/a5fc734b62180b2fe2c2cb025671dc71a64fc0b6/isic_archive/models/image.py#L66) for every image upon creation
* `readable` is not set upon creation, then is set to `True` or `False` in `ingestImage`
* `ingestImage` is not set upon creation, then is set [a `dict` of `None` values](https://github.com/ImageMarkup/isic-archive/blob/a5fc734b62180b2fe2c2cb025671dc71a64fc0b6/isic_archive/tasks/image.py#L91) in `ingestImage`
* `largeImage` is not set upon creation, then is set to [a `dict` of fully populated values](https://github.com/ImageMarkup/isic-archive/blob/a5fc734b62180b2fe2c2cb025671dc71a64fc0b6/isic_archive/tasks/image.py#L214) in `generateLargeImage`
* `superpixelsId` is not set upon creation, then is set to an ObjectId in `generateSuperpixels`

By just deleting all these fields here, we are rolling back the image to a state that it typically doesn't have in a normal pipeline. I'm not sure if this will cause any bugs until the image fields are fully re-populated. A more thorough approach is to initialize all of these to `None` / empty values upon image creation and simply reset them to that same state here, so we have assurance that the fields always exist.

isic_archive/models/image.py
```diff
@@ -40,6 +40,16 @@ def generateIsicId():
 
         return isicId
 
+    def renameImage(self, image, newName):
+        # assumes newName is of the format "ISIC_0123456"
+        oldName = image['name']
+        image['name'] = newName
+        for file_ in self.childFiles(image):
```
brianhelba, line 47: Maybe call this `childFile`, so we don't get further into name style issues.

isic_archive/models/image.py
```diff
@@ -20,6 +20,22 @@
...
+
+    imageDefaults = {
+        'ingested': False,
+        'readable': None,
+        'ingestionState': {
+            'largeImage': None,
+            'superpixelMask': None
+        },
+        'largeImage': {
+            'sourceName': None,
```
brianhelba, line 32:
```suggestion
            'expected': True,
            'sourceName': None,
```

PR #720, opened 2020-04-14
Properly assign roles to image files

danlamanna:
----
Needs migration.

### DIFF
--- a/isic_archive/models/image.py
+++ b/isic_archive/models/image.py
@@ -2,7 +2,7 @@ import itertools
 import mimetypes
 import os
 import secrets
-from typing import List, Tuple
+from typing import List, Optional, Tuple
 
 import numpy
 
@@ -95,13 +95,8 @@ class Image(Item):
         return image
 
     def resetImageForIngest(self, image):
-        originalFile = self.originalFile(image)
-        originalFileId = originalFile['_id'] if originalFile else None
-
         for childFile in self.childFiles(image):
-            if originalFileId and childFile['_id'] == originalFileId:
-                continue
-            else:
+            if childFile['imageRole'] != 'original':
                 File().remove(childFile)
 
         image.update(self.imageDefaults)
@@ -131,13 +126,11 @@ class Image(Item):
         # TODO: copy license from dataset to image
         return image
 
-    def originalFile(self, image):
-        original = File().findOne({
+    def originalFile(self, image) -> Optional[dict]:
+        return File().findOne({
             'itemId': image['_id'],
             'imageRole': 'original'
         })
-
-        return File().load(original['_id'], force=True)
 
     def strippedFile(self, image):
         return File().findOne({'itemId': image['_id'],
--- a/isic_archive/models/image.py
+++ b/isic_archive/models/image.py
@@ -56,6 +56,16 @@ class Image(Item):
 
         return isicId
 
+    def renameImage(self, image, newName):
+        # assumes newName is of the format "ISIC_0123456"
+        oldName = image['name']
+        image['name'] = newName
+        for childFile in self.childFiles(image):
+            childFile['name'] = childFile['name'].replace(oldName, newName)
+            File().save(childFile)
+
+        return Image().save(image)
+
     def createEmptyImage(self, originalFileRelpath, parentFolder, creator, dataset,
                          batch):
         newIsicId = self.generateIsicId()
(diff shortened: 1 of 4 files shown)


isic_archive/tasks/image.py
```diff
@@ -175,7 +177,7 @@ def generateSuperpixels(self, imageId):
 @app.task(bind=True)
 def generateLargeImage(self, imageId):
     imageItem = Image().load(id=imageId, force=True)
-    imageFile = Image().strippedFile(imageItem)
+    imageFile = Image().originalFile(imageItem)
```
danlamanna, line 180: I'm unsure about whether this change matters.

danlamanna:
After this migration we should run another which deletes all image files (not attachedTo) that have no role assigned to them.

isic_archive/models/image.py
```diff
@@ -92,11 +132,12 @@ def createImage(self, imageDataStream, imageDataSize, originalFileRelpath,
...
-            return File().load(image['largeImage']['originalId'], force=True)
-        else:
-            # Fallback if no large image metadata exists, but this isn't accurate on some old images
-            return Image().childFiles(image, limit=1, sort=[('created', SortDir.ASCENDING)])[0]
+        original = File().findOne({
+            'itemId': image['_id'],
+            'imageRole': 'original'
+        })
+
+        return Image().load(original['_id'], force=True)
```
brianhelba, line 140: This should just return the file.

isic_archive/tasks/zip.py
```diff
@@ -157,6 +157,10 @@ def ingestBatchFromZipfile(self, batchId):
                         logger.exception('An individual image failed to be extracted')
                         continue
 
+                    file_ = File().load(resp.json()['_id'], force=True)
```
brianhelba, line 160: `originalFile` would be more helpful.

scripts/migrateImageRoles.py
```diff
@@ -0,0 +1,80 @@
...
+        image['largeImage']['fileId'] if 'largeImage' in image else None
+    )
+    if large_image_file_id:
+        assert large_image_file_id not in taken_file_ids
+        taken_file_ids = taken_file_ids.union(set([large_image_file_id]))
+
+    original_files = [
+        f['_id']
+        for f in files
+        if re.match('^ISIC_\d{7}\.[^\.]+$', f['name'])  # noqa
```
brianhelba, line 49: Characters inside `[ ]` don't need to be escaped.

scripts/migrateImageRoles.py
```diff
@@ -0,0 +1,80 @@
...
+        and f['_id'] not in taken_file_ids
+    ]
+
+    # this happens with the VSHR dataset
+    if len(original_files) > 1:
+        print('warning: more than one original file, picking at random')
+    original_file_id = original_files[0]
```
brianhelba, line 56: Are all of them identical? Is the 0th arbitrary or always the correct one?

isic_archive/tasks/zip.py
danlamanna: This is the one place where that won't work because `imageRole` isn't set yet.

scripts/migrateImageRoles.py
danlamanna: I resolved this to fall back to the jpeg if they aren't identical. The only place this occurred was in a dataset with duplicate large images.

isic_archive/tasks/zip.py
brianhelba: Sorry, I was unclear. I just meant to change the name of this local variable.

isic_archive/models/image.py
```diff
@@ -91,12 +126,15 @@ def createImage(self, imageDataStream, imageDataSize, originalFileRelpath,
...
-        else:
-            # Fallback if no large image metadata exists, but this isn't accurate on some old images
-            return Image().childFiles(image, limit=1, sort=[('created', SortDir.ASCENDING)])[0]
+    def originalFile(self, image) -> Optional[dict]:
+        original = File().findOne({
+            'itemId': image['_id'],
+            'imageRole': 'original'
+        })
+
+        if original:
```
brianhelba, line 135: I still don't understand why this is happening. Isn't `original` simply the thing we want to return?

isic_archive/tasks/image.py
brianhelba: I think it's better this way, but I'm not critically worried about the old large_image files right now.

brianhelba: I think it's better this way, but I'm not critically worried about the old large_image files right now.

PR #723, opened 2020-04-21
Add ipython and mongo shell to vagrant

danlamanna:


ansible/vagrant-playbook.yml
```diff
@@ -65,10 +65,18 @@
         line: "127.0.0.1 minio"
       become: true
 
-    - name: Install dev requirements
+    - name: Install python dev tools
       pip:
-        name: tox
+        name: "{{ item }}"
```
brianhelba, line 70: `name` can be a list: https://docs.ansible.com/ansible/latest/modules/pip_module.html#parameter-name

That way it will install everything in a single pip invocation, rather than running it multiple times for each package.

ansible/vagrant-playbook.yml
```diff
@@ -65,10 +65,18 @@
...
         virtualenv: "{{ girder_virtualenv }}"
+      loop:
+        - "tox"
+        - "ipython"
```
brianhelba, line 74:
```suggestion
        name:
          - "tox"
          - "ipython"        
        virtualenv: "{{ girder_virtualenv }}"
```

PR #724, opened 2020-04-22
Remove extraneous image files migration

danlamanna:


brianhelba:
I'd like to talk through what this deletes, just in case we can find any other database faults in the process.

PR #725, opened 2020-04-22
Re-enable superpixels, fix issue with reset metadata

danlamanna:


isic_archive/models/image.py
```diff
@@ -108,6 +108,10 @@ def resetMetadata(self, image):
         image['meta']['clinical'] = {}
```
brianhelba, line 108: `image['meta']['acquisition']` has resolution properties we might want to keep.

isic_archive/models/image.py
```diff
@@ -108,6 +108,10 @@ def resetMetadata(self, image):
...
         image['meta']['unstructured'] = {}
         image['meta']['unstructuredExif'] = {}
+
+        if 'privateMeta' in image and 'age' in image['privateMeta']:
```
brianhelba, line 112: Is `privateMeta` ever not in `image`?

PR #745, opened 2020-12-23
Remove study endpoints

danlamanna:
@brianhelba Is this the sort of thing you were looking for? We could also leave the code and just remove the `.route` calls. Or we could just delete the endpoints entirely and they would 404 instead of 403.


danlamanna:
@brianhelba let me know if this is good to merge.

brianhelba (approved):

PR #751, opened 2021-03-16
Add index to create field

danlamanna:


isic_archive/api/image.py
```diff
@@ -119,6 +119,10 @@ def find(self, params):
         })
 
         filterFunc = Image().filter if detail else Image().filterSummary
+
+        if sort not in ['id', 'name', 'created']:
+            raise RestException('Sort may only be done on "id", "name", or "created" fields.')
```
brianhelba, line 124:
```suggestion
        if sort not in ['_id', 'name', 'created']:
            raise RestException('Sort may only be done on "_id", "name", or "created" fields.')
```

@danlamanna Right?

isic_archive/api/image.py
```diff
@@ -119,6 +119,10 @@ def find(self, params):
         })
 
         filterFunc = Image().filter if detail else Image().filterSummary
+
+        if sort not in ['_id', 'name', 'created']:
```
brianhelba, line 123: This should have been a set.

